Issue #3, opened 2023-08-16
PMP as an option for supervisor domain isolation


NicholasWoodIMG:
Add PMP as an option for supervisor domain isolation:

- Add PMP as a mode in table 3.1
- When PMP mode is used, then MTT is not required and PMP is used instead for physical memory isolation between domains
- When PMP mode is used, the SDID CSR can still be used to identify the currently active supervisor domain

This allows systems based on PMP or MTT as the underlying physical memory isolation mechanism to implement and support supervisor domains in a consistent way across the RVI architecture.

jackwdandrew:
Additionally, when only PMP and Bare modes are supported, mttp.MTTPN is redundant. Therefore an implementation that  can tie it to 0. Probably calling mttp.MTTPN out as WARL is sufficient.

rsahita:
Makes sense. Sub-fields of that CSR are already specified as WARL.

from 9-19 discussion - Krste's comment - will need to use SFENCE to flush TLBs with any PMP settings changes

kasanovic:
To clarify, if we did add SDID we may be able to provide more targeted SFENCE versions to reduce the microarchitectural translation/protection cache flushing  when PMP entries are swapped.

NicholasWoodIMG:
Thanks. That looks like it would be useful. Could we address that here in this document?

Another useful extension to PMP for supervisor domains would be to allow “switchable” PMP configurations indexed by SDID to avoid PMP reconfiguration on context switch. Similar to MMU configurations indexed by VMID.


From: Krste Asanovic ***@***.***>
Sent: Tuesday, September 19, 2023 3:28 PM
To: riscv/riscv-smmtt ***@***.***>
Cc: Nicholas Wood ***@***.***>; Author ***@***.***>
Subject: [EXTERNAL] Re: [riscv/riscv-smmtt] PMP as an option for supervisor domain isolation (Issue #3)

*** CAUTION: This email originates from a source not known to Imagination Technologies. Think before you click a link or open an attachment ***



To clarify, if we did add SDID we may be able to provide more targeted SFENCE versions to reduce the microarchitectural translation/protection cache flushing when PMP entries are swapped.

rsahita:
Eckhard: Do we also need an informative section for the "no-MTT" Scenario where IOPMP is used for Access Control (instead of IOMMU-MTT)?

NicholasWoodIMG:
Table 2. Encoding of mttMODE field:

[...]

PR #19, opened 2024-01-22
Addressing issue #3

rsahita:
PR to address issue [3](https://github.com/riscv/riscv-smmtt/issues/3)

@NicholasWoodIMG

### DIFF
--- a/chapter2.adoc
+++ b/chapter2.adoc
@@ -13,7 +13,7 @@ instructions  - herewith called the `supervisor domain identifier` assigned to
 the hart. Supervisor domains are orthogonal to hart privilege levels and since
 Smmtt enables physical memory isolation, there is one CSR (per hart) managed by
 M-mode. `Smsdid` is expected to be used in conjunction with `Smmtt` for physical
-memory isolation along with mechanisms such as `e(PMP)`. Device side
+memory isolation along with mechanisms such as `PMP` and `Smepmp`. Device side
 accesses are addressed in the `IO-MTT` extension. Isolation of data within a
 device is out of scope of this specification.
 
@@ -31,11 +31,9 @@ the RDSM (except for when explicitly allowed by the RDSM). Privilege levels may
 affect changes in the MTT under purview of the Supervisor Domain Security
 Manager (SDSM) either through an SBI interface into M-mode (or may have the
 ability to edit MTT structures by virtue of how the MTT structure in memory is
-accessible to lower privilege levels). Other memory access control mechanisms
-may also be used to enforce supervisor domain isolation for use cases that do
-not require the full flexibility of MTT. For example, (e)PMP configurations
-controlled by M-mode during domain context switching. In this case the MTT-mode
-`Bare` may be used. When MTT mode is `Bare`, MTT is not enforced, but the SDID
+accessible to lower privilege levels). MTT and e(PMP) are always active.
+MTT may be configured to be `Bare` if granular memory access control
+is not required. The SDID
 CSR defined by `Smsdid` is used as defined.
 
 * `Svpams` (<<Svpams>>) - Physical address metadata selector - When access to
--- a/chapter2.adoc
+++ b/chapter2.adoc
@@ -12,8 +12,10 @@ that can be held in an M-mode CSR and modifiable by the RDSM via CSR r/w
 instructions  - herewith called the `supervisor domain identifier` assigned to
 the hart. Supervisor domains are orthogonal to hart privilege levels and since
 Smmtt enables physical memory isolation, there is one CSR (per hart) managed by
-M-mode. Device side accesses are addressed in the `IO-MTT` extension. Isolation
-of data within a device is out of scope of this specification.
+M-mode. `Smsdid` is expected to be used in conjunction with `Smmtt` for physical
+memory isolation along with mechanisms such as `e(PMP)`. Device side
+accesses are addressed in the `IO-MTT` extension. Isolation of data within a
+device is out of scope of this specification.
 
 * `Smmtt` (<<Smmtt>>) - An interface to set the access permissions for a memory
 region or page associated with a supervisor domain. This interface allows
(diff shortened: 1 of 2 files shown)


chapter2.adoc
```diff
@@ -29,7 +31,12 @@ the RDSM (except for when explicitly allowed by the RDSM). Privilege levels may
 affect changes in the MTT under purview of the Supervisor Domain Security
 Manager (SDSM) either through an SBI interface into M-mode (or may have the
 ability to edit MTT structures by virtue of how the MTT structure in memory is
-accessible to lower privilege levels).
+accessible to lower privilege levels). Other memory access control mechanisms
+may also be used to enforce supervisor domain isolation for use cases that do
+not require the full flexibility of MTT. For example, (e)PMP configurations
+controlled by M-mode during domain context switching. In this case the MTT-mode
+`Other` may be used. When MTT mode is `Other`, MTT is not used, but the SDID
+CSR defined by `Smsdid` is used as defined.
```
ved-rivos, line 39: It is better to define that there is no option to disable memory protection provided by MTT but the MTT enforcement can be effectively disabled by setting `mode` to `Bare`. When MTT is in `Bare`, mode it does not enforce any memory protection. With this definition we do not need a `Other` encoding for MTT. It also does not require this text which would imply that e(PMP) are mutually exclusive with MTT whereas e(PMP) are not bypassed even with MTT is not `Bare`.

NicholasWoodIMG: Yes, this was my understanding following other tweaks made to the spec.

Basically "MTT mode" turns MTT on or off. All we need to say is that when MTT mode is "bare" (no MTT enforcement), other architectural memory isolation features can still be used to enforce physical memory isolation. 

And all the other parts of smmtt can still be used as well - SDID, interrupt extensions, debug extensions, etc. Only difference is MTT is disabled/not supported.

I don't know if IOMTT makes sense without MTT...

ved-rivos: The `mttp` CSR must always exists. The `mode` field of `mttp` is WARL so it is legal to hard wire the `mode`to `Bare`.  There are two functions provided by IOMTT:
- Supervisor domain classifier - classify a DMA request to a SDID (and MTT)
- MTT check
The supervisor domain classification is used to identify the IOMMU that should govern the DMA requests. So similar will likely apply that the MTT mode in IOMTT may be hard wired to `Bare` but the classifier to classify DMA requests to supervisor domains would exist.

rsahita: ack - if some other mechanism is defined in the future, such an "Other" option can be added at that point - so I will delete the change patch for the mttp.MODE CSR field and update the text above accordingly to specify use of 'Bare' mode.

rsahita (commented):
@ved-rivos @NicholasWoodIMG see updated

chapter2.adoc
```diff
@@ -12,8 +12,10 @@ that can be held in an M-mode CSR and modifiable by the RDSM via CSR r/w
 instructions  - herewith called the `supervisor domain identifier` assigned to
 the hart. Supervisor domains are orthogonal to hart privilege levels and since
 Smmtt enables physical memory isolation, there is one CSR (per hart) managed by
-M-mode. Device side accesses are addressed in the `IO-MTT` extension. Isolation
-of data within a device is out of scope of this specification.
+M-mode. `Smsdid` is expected to be used in conjunction with `Smmtt` for physical
+memory isolation along with alternate mechanisms such as `e(PMP)`. Device side
+accesses are addressed in the `IO-MTT` extension. Isolation of data within a
+device is out of scope of this specification.
```
ved-rivos, line 18: The text seems to imply that e(PMP) are alternates to MTT. Whereas MTT and e(PMP) are applicable all the time.

chapter2.adoc
```diff
@@ -29,7 +31,12 @@ the RDSM (except for when explicitly allowed by the RDSM). Privilege levels may
 affect changes in the MTT under purview of the Supervisor Domain Security
 Manager (SDSM) either through an SBI interface into M-mode (or may have the
 ability to edit MTT structures by virtue of how the MTT structure in memory is
-accessible to lower privilege levels).
+accessible to lower privilege levels). Other memory access control mechanisms
+may also be used to enforce supervisor domain isolation for use cases that do
+not require the full flexibility of MTT. For example, (e)PMP configurations
+controlled by M-mode during domain context switching. In this case the MTT-mode
+`Bare` may be used. When MTT mode is `Bare`, MTT is not enforced, but the SDID
```
ved-rivos, line 38: I think we should state:
1. MTT and e(PMP) are always active
2. MTT may be configured to be Bare if granular memory access control is not required
Then we dont need to say anything about SDID etc. MTT is always active - it just happens to be Bare.

gagachang:
I suggest using `Smepmp` instead of `ePMP`, since the name of extension is `Smepmp`.

chapter2.adoc
NicholasWoodIMG: I can't see the latest commits for some reason?

But I think I agree with Ved's reasoning. All physical checks are always there. MTT mode "bare" simply disables the MTT check (always passes).

Then we can leave use case discussions to the security model. We already have both with and without MTT there.

Nit - it should be (e)PMP, not e(PMP)? Or PMP/ePMP to be really clear... It could be PMA as well, or whatever. 

But all we need to say here is that all RISC-V physical memory checks are always performed, whether MTT is active or not.

chapter2.adoc
```diff
@@ -12,8 +12,10 @@ that can be held in an M-mode CSR and modifiable by the RDSM via CSR r/w
 instructions  - herewith called the `supervisor domain identifier` assigned to
 the hart. Supervisor domains are orthogonal to hart privilege levels and since
 Smmtt enables physical memory isolation, there is one CSR (per hart) managed by
-M-mode. Device side accesses are addressed in the `IO-MTT` extension. Isolation
-of data within a device is out of scope of this specification.
+M-mode. `Smsdid` is expected to be used in conjunction with `Smmtt` for physical
+memory isolation along with mechanisms such as `e(PMP)`. Device side
```
rsahita, line 16:
```suggestion
memory isolation along with mechanisms such as `PMP` and `Smepmp`. Device side
```

chapter2.adoc
rsahita, line 38:
```suggestion
accessible to lower privilege levels). MTT and e(PMP) are always active.
MTT may be configured to be `Bare` if granular memory access control
is not required. The SDID
```

PR #35, opened 2024-02-23
Updates to Smsdedbg per issue 21 and 26

rsahita:
add changes to Smsdedbg normative behavior per discussion in issue 26.
Also proposed changes to be documented in the Secure Ext debug spec to enable the behavior required for Smsdedbg.

### DIFF
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -140,10 +140,8 @@ configuration for supervisor domains:
 
 . `Smsdia` uses `msdcfg.SDICN` to specify the active configuration for
   the supervisor domain interrupt controller associated with the hart.
-. External-debug-allowed state for a supervisor domain is managed via the
-  `msdcfg.sdedbgalw` bit.
-. External-trace-allowed state for a supervisor domain is managed via the
-  `msdcfg.sdetrcalw` bit.
+. `Smsdedbg` specifies the `msdcfg.sdedbgalw` bit to manage external-debug for a supervisor domain.
+. `Smsdetrc` specifies the `msdcfg.sdetrcalw` bit to manage external-trace for a supervisor domain.
 . `Smqosid` specifies the control bits `SSM`, `SRL`, `SML` and `SQRID` to enable
   the RDSM to manage QoS controls for supervisor domains.
 
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -134,12 +134,21 @@ after writing `mttp`.
 
 The `msdcfg` is a 32-bit read/write register, formatted as shown in <<MSDCFG>>.
 This CSR is used by M-mode software to specify the active configuration for
-capabilities of the supervisor domain when associated with a hart. The
-extension `Smsdia` uses `msdcfg.SDICN` to specify the active configuration for
-the supervisor domain interrupt controller associated with the hart. Similarly
-the external debug allowed state for a supervisor domain is controlled via the
-`msdcfg.SDEDBALW` bit. Details of `Smsdia` and `Smsdedbg` are described in
-their respective sections in this specification.
+capabilities of the supervisor domain when associated with a hart. +
+The following extensions use the `msdcfg` register to specify additional
+configuration for supervisor domains:
+
+. `Smsdia` uses `msdcfg.SDICN` to specify the active configuration for
+  the supervisor domain interrupt controller associated with the hart.
+. External-debug-allowed state for a supervisor domain is managed via the
+  `msdcfg.sdedbgalw` bit.
+. External-trace-allowed state for a supervisor domain is managed via the
+  `msdcfg.sdetrcalw` bit.
+. `Smqosid` specifies the control bits `SSM`, `SRL`, `SML` and `SQRID` to enable
+  the RDSM to manage QoS controls for supervisor domains.
+
+Details of `Smsdia`, `Smsdedbg`, `Smsdetrc` and `Smqosid` are described in their
+respective sections in this specification.
 
 [[MSDCFG]]
 .`msdcfg` register
@@ -148,8 +157,12 @@ their respective sections in this specification.
 ....
 {reg: [
   {bits:  6, name: 'SDICN'},
-  {bits:  1, name: 'WPRI'},
+  {bits:  1, name: 'SSM'},
   {bits:  1, name: 'SDEDBGALW'},
-  {bits: 25, name: 'WPRI'},
+  {bits:  1, name: 'SDETRCALW'},
+  {bits:  11, name: 'WPRI'},
+  {bits:  4, name: 'SRL'},
+  {bits:  4, name: 'SML'},
+  {bits:  4, name: 'SQRID'},
 ], config:{lanes: 4, hspace:1024}}
 ....
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -140,8 +140,10 @@ configuration for supervisor domains:
 
 . `Smsdia` uses `msdcfg.SDICN` to specify the active configuration for
   the supervisor domain interrupt controller associated with the hart.
-. `Smsdedbg` specifies the `msdcfg.sdedbgalw` bit to manage external-debug for a supervisor domain.
-. `Smsdetrc` specifies the `msdcfg.sdetrcalw` bit to manage external-trace for a supervisor domain.
+. `Smsdedbg` specifies the `msdcfg.sdedbgalw` bit to manage 
+  external-debug for a supervisor domain.
+. `Smsdetrc` specifies the `msdcfg.sdetrcalw` bit to manage 
+  external-trace for a supervisor domain.
 . `Smqosid` specifies the control bits `SSM`, `SRL`, `SML` and `SQRID` to enable
   the RDSM to manage QoS controls for supervisor domains.
 
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -118,7 +118,7 @@ The `metrcen` is an enable control for external trace for the `M-mode` driven by
 an external trace module and is expected to be established by the RoT (following
 RISC-V security model recommendations SR_GEN_007 and 012). When privilege is
 `M-mode`, the `metrcen` determines the state of the `halted` side band signal sent by the hart to the
-trace encoder. Per the Efficient trace specification cite:[Etrc], this side-band
+trace encoder. Per the Efficient trace specification cite:[ETrc], this side-band
 `halted` signal being asserted, stops subsequent tracing from the hart. On this
 signal being deasserted, the encoder can start tracing again.
 
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -6,22 +6,58 @@ The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
 current scheduled SD is allowed to be external-debugged. This bit is context
 switched (along with rest of the `msdcfg`) per SD.
 
+When M-mode external debug is enabled, all supervisor domains may also be
+debugged by an external debugger irrespective of the configuration held in
+`msdcfg.SDEDBGALW`.
+
+When M-mode external debug is disabled, whether execution at privilege modes
+less than `M-mode` may be debugged by an external debugger depends on the
+configuration held in `msdcfg.SDEDBGALW`.
+
+When `msdcfg.SDEDBGALW` = 0, external debug is disallowed. Abstract commands
+and halt request from the debug module are suppressed and stay pending.
+
+When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
+`M-mode` is allowed, and:
+
+* A halt request may transition the hart to Debug Mode.
+* Abstract commands and program buffer execution can access state of privilege
+modes less than `M-mode`.
+* Read and Write of `Sdtrig` CSRs is allowed.
+* Debugger memory accesses occur with either `S-mode` or `U-mode` privilege (as
+if `aamvirtual` = 1 and `MPP` != `M-mode`).
+
 === `Smsdedbg` interaction with external debug security controls (Informative)
 
 [caption="Figure {counter:image}: ", reftext="Figure {image}"]
 [title= "External Debug for Supervisor Domain", id=Smsdedbg_img]
 image::images/Smsdedbg.png[]
 
-This section will be moved into the non-ISA specification for external debug
-security. It is described here as informational.
+This section will be moved into the specification for external debug security.
+It is described in this specification as informational.
 
 The `medbgen` is an enable control for external debug for the M-mode driven by
 the debug module and is expected to be established by the RoT (following RISC-V
 Security Model recommendation SR_GEN_007 and SR_GEN_012). When privilege is `M`,
 the `medbgen` gates the `haltreq` from the debug module and if is 0 prevents
-the hart from entering external debug mode. When privilege is less than `M`, the
-OR of the `MSDCFG.sdedbgalw` and `medbgen` gates the `haltreq` from the debug
-module and the hart will enter external debug mode if either is 1.
+the hart from entering external debug mode.
+
+The following change is proposed to behavior of `M-mode` access to triggers with
+`dmode` = 1. This change allows the RDSM to remain in control of external debug
+for supervisor domains (unless the RDSM is itself under external debug).
+
+When `medbgen` is 0 and privilege is `M-mode`:
+
+* M-mode can read and write triggers, *including triggers with `dmode` = 1
+without restrictions*.
+* Abstract commands and halt request from external debugger stay pending while
+privilege is `M-mode`.
+* All trigger-matching is suppressed (similar to how `MIE` or `MTE` would
+suppress them)
+
+When privilege is less than `M`, the OR of the `MSDCFG.sdedbgalw` and `medbgen`
+gates the `haltreq` from the debug module and the hart will enter external debug
+mode if either is 1.
 
 The configuration for `MSDCFG.sdedbgalw` may be obtained from the manifest/
 configuration of the supervisoer domain and should be managed by the M-mode root
@@ -30,8 +66,8 @@ security manager using secure memory.
 When `medbgen` is 1, there are no restrictions. When `medbgen` is 0 but
 `MSDCFG.sdedbgalw` is 1, then the external debug mode may be entered but is
 restricted by `M-mode` software to prevent privileged CSR accesses and memory
-accesses by instructions executed in external debug mode cannot use `M` privilege.
-When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the M-mode root domain
-security manager must not configure triggers with `action=1`. Triggers for the
-supervisor domain are expected to be controlled by the root domain security
-manager to prevent any leakage of information.
+accesses by instructions executed in external debug mode cannot use `M-mode`
+privilege. When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the `M-mode`
+root domain security manager must not configure triggers with `action=1`.
+Triggers for the supervisor domain are expected to be controlled by the root
+domain security manager to prevent any leakage of information.
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -73,7 +73,8 @@ accesses by instructions executed in Debug Mode cannot use `M-mode`
 privilege. When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the `M-mode`
 root domain security manager must not configure triggers with `action=1`.
 Triggers for the supervisor domain must be controlled by the root
-domain security manager to prevent any leakage of information across security domains.
+domain security manager to prevent any leakage of information across security
+domains.
 
 === `Smsdetrc`: External Trace allowed control for Supervisor Domain
 
@@ -102,12 +103,27 @@ It is described in this specification as informational.
 The `metrcen` is an enable control for external trace for the `M-mode` driven by
 an external trace module and is expected to be established by the RoT (following
 RISC-V security model recommendations SR_GEN_007 and 012). When privilege is
-`M-mode`, the `metrcen` gates the `trTEActive` control into the hart trace
-encoder and if 0, prevents the hart from generating any trace output.
+`M-mode`, the `metrcen` gates the `halted` signal (from the hart) into the hart
+trace encoder. Per the specification <<R24>>, this side-band `halted` signal
+being asserted, prevents the hart from generating any trace output.
 
-When `metrcen` is 0 and privilege is `M-mode`, `M-mode` activity cannot be
-traced and `M-mode` software can enforce that `trTeEnable` is set to the state
-of `msdcfg.SDETRCALW` for the SD it context switches the hart into.
+[NOTE]
+====
+Implementation of the `halted` side-band signal is required to support external
+tracing with supervisor domains
+====
 
-When `metrcen` is 1, there are no restrictions and all privilege modes
-(inclusive of `M-mode` and any SDs managed by the RDSM) are traceable.
+When `metrcen` is 0:
+`halted` is asserted on:
+
+* entry to Debug mode
+* entry to `M-mode` privilege
+* entry to privilege less than `M-mode` with `msdcfg.SDETRCALW` = 0.
+
+and, `halted` is deasserted on entry to privilege less than `M-mode` with
+`msdcfg.SDETRCALW` = 0
+
+When `metrcen` is 1:
+There are no restrictions and all privilege modes (inclusive of `M-mode` and any
+SDs managed by the RDSM) are traceable. In this case, `halted` is asserted on
+entry to Debug mode, and deasserted on resumption from Debug Mode.
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -50,17 +50,17 @@ When `msdcfg.sdedbgalw` = 1 then external debug of privilege modes less than
 `M-mode` is allowed for such a supervisor domain on a per-hart basis, 
 with the additional requirements listed below. +
 
-To enforce the above controls specified by this extension, the following
-functional and security requirements must be met by the external secure debug
-system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
-isolation:
-
-. External debug of M-mode can be enabled only by the HW RoT of the RDSM
- (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
 . External debug must be able to access supervisor domain memory and/or state. 
   In this context, "state" includes all supervisor domain resources accessible per the 
   Debug specification cite:[ExtDbg].
 . Entry to Debug Mode from a supervisor domain is allowed.
+
+To enforce the above controls specified by this extension, the following
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec]:
+
+. External debug of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
 . The RDSM must be able to context switch all triggers even when `d-mode` = 1.
   This functional change allows the RDSM to remain in control of external debug
   for supervisor domains when it is not under external debug itself.
@@ -81,10 +81,14 @@ When `msdcfg.sdetrcalw` = 1 then external trace of privilege modes less than
 `M-mode` shall be allowed for the SD on a per hart basis, with the
 additional requirements listed below.
 
+. When external tracing is disabled, tracing must be stopped on:
+.. entry to Debug mode
+.. entry to `M-mode` privilege
+.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
+
 To enforce the above controls specified by this extension, the following
-functional and security requirements must be met by the external secure debug
-system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
-isolation:
+functional and security requirements must be met by the external debug security
+system cite:[ExtDbgSec]:
 
 . External trace of M-mode can be enabled only by the HW RoT of the RDSM
  (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -40,13 +40,11 @@ When M-mode external debug is disabled, whether execution at privilege modes
 less than `M-mode` may be debugged by an external debugger depends on the
 configuration held in `msdcfg.sdedbgalw`, as described below:
 
-When `msdcfg.sdedbgalw` = 0, external debug is disallowed for the supervisor
-domain. Abstract commands and halt request from the debug module are suppressed
-and stay pending while the supervisor domain is active, on a per-hart basis.
-Other debug operations (current or defined in the future) must similarly be kept
-pending while the supervisor domain is active on a hart. Triggers for the
-supervisor domain must be controlled by the RDSM to prevent any leakage of
-information across security domains.
+When `msdcfg.sdedbgalw` is 0:
+
+* Access by external debuggers to the memory and/or state of the supervisor domain is disallowed.
+* Entry to Debug Mode from a supervisor domain is disallowed.
+
 
 When `msdcfg.sdedbgalw` = 1 then external debug of privilege modes less than
 `M-mode` is allowed for such a supervisor domain on a per-hart basis. +
@@ -56,10 +54,8 @@ functional and security requirements must be met by the external secure debug
 system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
 isolation:
 
-. The enable control for external debug driven by the external debug module is
-  expected to be established by the platform root-of-trust following RISC-V
-  Security Model recommendations SR_GEN_007 and SR_GEN_012. When the control is
-  in cleared state, the hart should not be able to enter external debug.
+. External debug of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
 . External debug must be able to transition the hart to Debug Mode and access
   supervisor domain memory and state. In this context, "state" includes all
   non-M-mode resources accessible per the Debug specification cite:[ExtDbg].
@@ -68,14 +64,6 @@ isolation:
 . The RDSM must be able to context switch all triggers even when `d-mode` = 1.
   This functional change allows the RDSM to remain in control of external debug
   for supervisor domains when it is not under external debug itself.
-
-The following figure is non-normative and is intended to illustrate the use of
-the `msdcfg.sdedbgalw` control - the normative specification will be in the
-specification for external debug security cite:[ExtDbgSec].
-
-[caption="Figure {counter:image}: ", reftext="Figure {image}"]
-[title= "External Debug opt-in for Supervisor Domain", id=Smsdedbg_img]
-image::images/Smsdedbg.png[]
 
 === `Smsdetrc`: External Trace allowed control for Supervisor Domain
 
@@ -99,10 +87,8 @@ functional and security requirements must be met by the external secure debug
 system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
 isolation:
 
-. The enable control for external trace driven by the external trace module is
-  expected to be established by the platform root-of-trust following RISC-V
-  Security Model recommendations SR_GEN_007 and SR_GEN_012. When the control is
-  in cleared state, the hart should not be able to generate trace.
+. External trace of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
 . Per the Efficient trace specification cite:[ETrc], the side-band `halted`
   signal being asserted, stops subsequent tracing from the hart. On this signal
   being deasserted, the encoder can start tracing again. Implementation of this
@@ -118,7 +104,3 @@ isolation:
   `halted`) on:
 .. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0
 
-When external tracing is enabled (and authorized), there are no restrictions and
-all privilege modes (inclusive of `M-mode` and any supervisor domains managed by
-the RDSM) are traceable. In this case, tracing shall be enabled (via asserting
-`halted`) on entry to Debug mode, and stopped on resumption from Debug Mode.
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -56,11 +56,10 @@ isolation:
 
 . External debug of M-mode can be enabled only by the HW RoT of the RDSM
  (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
-. External debug must be able to transition the hart to Debug Mode and access
-  supervisor domain memory and state. In this context, "state" includes all
-  non-M-mode resources accessible per the Debug specification cite:[ExtDbg].
-. Triggers must match and fire on a hart only when a supervisor domain with this
-  configuration is active on a hart (and not in M-mode).
+. External debug must be able to access supervisor domain memory and/or state. 
+  In this context, "state" includes all supervisor domain resources accessible per the 
+  Debug specification cite:[ExtDbg].
+. Entry to Debug Mode from a supervisor domain is allowed.
 . The RDSM must be able to context switch all triggers even when `d-mode` = 1.
   This functional change allows the RDSM to remain in control of external debug
   for supervisor domains when it is not under external debug itself.
@@ -87,12 +86,10 @@ isolation:
 
 . External trace of M-mode can be enabled only by the HW RoT of the RDSM
  (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
-. When external tracing is disabled, tracing should be stopped (via asserting
-  `halted`) on:
+. When external tracing is disabled, tracing must be stopped on:
 .. entry to Debug mode
 .. entry to `M-mode` privilege
 .. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
-. When extracing tracing is disabled, tracing should be started (via deasserting
-  `halted`) on:
-.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0
+. When external tracing is disabled, tracing must be started on:
+.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 1
 
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -75,9 +75,7 @@ When M-mode external trace is disabled, whether execution at privilege modes
 less than `M-mode` may be traced by an external trace tool depends on the
 configuration held in `msdcfg.sdetrcalw`, as described below:
 
-When `msdcfg.sdetrcalw` = 0, external trace is disallowed for the SD. All trace
-output from the trace encoder module are disabled while the SD is active on a
-per-hart basis.
+When `msdcfg.sdetrcalw` = 0, external trace of the supervisor domain is disallowed.
 
 When `msdcfg.sdetrcalw` = 1 then external trace of privilege modes less than
 `M-mode` shall be allowed for the SD on a per hart basis.
@@ -89,12 +87,6 @@ isolation:
 
 . External trace of M-mode can be enabled only by the HW RoT of the RDSM
  (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
-. Per the Efficient trace specification cite:[ETrc], the side-band `halted`
-  signal being asserted, stops subsequent tracing from the hart. On this signal
-  being deasserted, the encoder can start tracing again. Implementation of this
-  `halted` side-band signal is required to support external tracing with
-  supervisor domains. The state of the `halted` side-band signal shall be gated
-  by the external tracing enable control for M-mode.
 . When external tracing is disabled, tracing should be stopped (via asserting
   `halted`) on:
 .. entry to Debug mode
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -104,14 +104,17 @@ The `metrcen` is an enable control for external trace for the `M-mode` driven by
 an external trace module and is expected to be established by the RoT (following
 RISC-V security model recommendations SR_GEN_007 and 012). When privilege is
 `M-mode`, the `metrcen` gates the `halted` signal (from the hart) into the hart
-trace encoder. Per the specification <<R24>>, this side-band `halted` signal
-being asserted, prevents the hart from generating any trace output.
+trace encoder. Per the Efficient trace specification cite:[Etrc], this side-band
+`halted` signal being asserted, stops subsequent tracing from the hart. On this
+signal being deasserted, the encode can start tracing again.
 
 [NOTE]
 ====
 Implementation of the `halted` side-band signal is required to support external
 tracing with supervisor domains
 ====
+
+`Smsdetrc` specifies the following behavior for tracing with supervisor domains:
 
 When `metrcen` is 0:
 `halted` is asserted on:
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -9,6 +9,11 @@ software running on a separate target RISC-V platform, via a trace control/data
 transport that provides external access to the trace encoder controls
 cite:[ETrc]. This extension only affects debug or trace orchestrated by an
 external actor. Self-hosted debug (and trace) are managed by the RDSM.
+This sub-specification refers to the Debug specification cite:[ExtDbg],
+Trace specification cite:[ETrc], and the External debug security extension
+cite:[ExtDbgSec] for the description and behavior of controls outside
+the scope of this specification, but which interact with controls
+specified in this specification when supervisor domains are used.
 
 === `Smsdedbg`: External Debug allowed control for Supervisor Domain
 
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -39,7 +39,8 @@ modes less than `M-mode`. In this context, "state" includes all resources
 accessible per the Debug specification cite:[ExtDbg].
 * Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
 * Debugger memory accesses may occur with `S-mode` or `U-mode` privilege per
-software execution in the SD, and as if `aamvirtual` = 1 and `MPP` != `M-mode`.
+software execution in the SD, and as if `aamvirtual` = 1 and `MPP` != `M-mode`
+and `relaxedpriv`=0.
 
 ==== `Smsdedbg` interaction with external debug security controls (Informative)
 
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -27,7 +27,7 @@ When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
 * A halt request may transition the hart to Debug Mode.
 * Abstract commands and program buffer execution can access state of privilege
 modes less than `M-mode`.
-* Read and Write of `Sdtrig` CSRs is allowed.
+* Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
 * Debugger memory accesses occur with either `S-mode` or `U-mode` privilege (as
 if `aamvirtual` = 1 and `MPP` != `M-mode`).
 
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -29,8 +29,11 @@ a hart.
 When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
 `M-mode` is allowed for the SD on a per-hart basis, and:
 
-* A valid request e.g. `haltreq` may transition the hart to Debug Mode. See
-other cases in RISC-V Debug specification cite:[ExtDbg].
+* A valid request (e.g. `halt`) or a trigger matching/firing may transition the hart to
+Debug Mode. See other cases in RISC-V Debug specification cite:[ExtDbg]. 
+Note that the `resethaltreq` is not applicable since it can be done only
+following a reset and the hart cannot be at privilege mode less than
+`M-mode` with `msdcfg.SDEDBGALW`=1 at de-assertion of reset.
 * Abstract commands and program buffer execution can access state of privilege
 modes less than `M-mode`. In this context, "state" includes all resources
 accessible per the Debug specification cite:[ExtDbg].
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -69,7 +69,12 @@ without restrictions*.
 * Abstract commands and requests from an external debugger, that transition the
 hart into a halted state, stay pending while privilege is `M-mode` cite:[ExtDbg].
 * All trigger-matching is prevented from matching or firing, similar to how
-`MIE` or `MTE` would behave.
+`MIE` or `MTE` would behave. This behavior is specifically as stated in
+section 5.4 of cite:[ExtDbg]: While `MIE` in `mstatus` is 0, the hardware
+prevents triggers with `action`=0 from matching or firing while in `M-mode` .
+Also from section 5.7.6 of cite:[ExtDbg], M-mode trigger enable `mte`=0
+(disabled) specifies that triggers with `action`=0 do not match/fire
+while the hart is in M-mode.
 
 When privilege is less than `M`, the OR of the `MSDCFG.sdedbgalw` and `medbgen`
 gates the `haltreq` from the debug module and the hart will enter Debug
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -92,8 +92,3 @@ system cite:[ExtDbgSec]:
 
 . External trace of M-mode can be enabled only by the HW RoT of the RDSM
  (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
-. When external tracing is disabled, tracing must be stopped on:
-.. entry to Debug mode
-.. entry to `M-mode` privilege
-.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
-
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -92,6 +92,4 @@ isolation:
 .. entry to Debug mode
 .. entry to `M-mode` privilege
 .. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
-. When external tracing is disabled, tracing must be started on:
-.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 1
 
--- a/chapter8.adoc
+++ b/chapter8.adoc
@@ -2,161 +2,123 @@
 [[Smsdedbg]]
 == Supervisor Domain External Trace and Debug
 
-The <<MSDCFG>> CSR contains the `sdedbgalw` and `sdetrcalw` bits that control
-if the current scheduled SD is allowed to be external-debugged or to be
-external-traced cite:[ExtDbg], respectively. External trace refers to tracing
-software running on a separate target RISC-V platform, via a trace control/data
-transport that provides external access to the trace encoder controls
-cite:[ETrc]. This extension only affects debug or trace orchestrated by an
-external actor. Self-hosted debug (and trace) are managed by the RDSM.
-This sub-specification refers to the Debug specification cite:[ExtDbg],
-Trace specification cite:[ETrc], and the External debug security extension
-cite:[ExtDbgSec] for the description and behavior of controls outside
-the scope of this specification, but which interact with controls
-specified in this specification when supervisor domains are used.
+This chapter describes two extensions `Smsdedbg` and `Smsdetrc` that enable a
+RDSM to manage isolation of supervisor domains when external debug and trace
+mechanisms are being utilized. `Smsdedbg` defines the `sdedbgalw` bit in the
+<<MSDCFG>> CSR that specifies if a supervisor domain is allowed to be
+externally-debugged. Similarly, `Smsdetrc` defines the `sdetrcalw` bit in the
+<<MSDCFG>> CSR that specifies if a supervisor domain is allowed to be
+externally-traced cite:[ExtDbg]. External trace refers to tracing software
+running on a separate target RISC-V platform, via a trace control/data transport
+that provides external access to the trace encoder controls cite:[ETrc]. These
+two extensions only affect debug and trace orchestrated by an external actor -
+self-hosted debug (and trace) are managed directly by the RDSM. This chapter
+refers to the Debug specification cite:[ExtDbg], Trace specification
+cite:[ETrc], and the External debug security extension cite:[ExtDbgSec] for the
+description and behavior of controls outside the scope of this specification,
+but which interact with controls specified in this specification when one or
+more supervisor domains are used. Note that `Smsdedbg` and `Smsdetrc` may be
+used as independent extensions even when a single supervisor domain is being
+utilized on a platform.
+
+[NOTE]
+====
+The configuration for `msdcfg.sdedbgalw` and `msdcfg.sdetrcalw` is expected to
+be obtained from the manifest/configuration of the supervisor domain and
+reported in the attestation report for the supervisor domain and it's workloads.
+The RDSM should manage this supervisor domain configuration state for context
+switching using secure memory.
+====
 
 === `Smsdedbg`: External Debug allowed control for Supervisor Domain
 
 When M-mode external debug is enabled, all supervisor domains may also be
 debugged by an external debugger irrespective of the configuration held in
-`msdcfg.SDEDBGALW`.
+`msdcfg.sdedbgalw`.
 
 When M-mode external debug is disabled, whether execution at privilege modes
 less than `M-mode` may be debugged by an external debugger depends on the
-configuration held in `msdcfg.SDEDBGALW`.
+configuration held in `msdcfg.sdedbgalw`, as described below:
 
-When `msdcfg.SDEDBGALW` = 0, external debug is disallowed for the SD. Abstract
-commands and halt request from the debug module are suppressed and stay pending
-while the SD is active, on a per-hart basis. Other debug operations (current or
-defined in the future) must similarly be kept pending while the SD is active on
-a hart.
+When `msdcfg.sdedbgalw` = 0, external debug is disallowed for the supervisor
+domain. Abstract commands and halt request from the debug module are suppressed
+and stay pending while the supervisor domain is active, on a per-hart basis.
+Other debug operations (current or defined in the future) must similarly be kept
+pending while the supervisor domain is active on a hart. Triggers for the
+supervisor domain must be controlled by the RDSM to prevent any leakage of
+information across security domains.
 
-When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
-`M-mode` is allowed for the SD on a per-hart basis, and:
+When `msdcfg.sdedbgalw` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for such a supervisor domain on a per-hart basis. +
 
-* A valid request (e.g. `halt`) or a trigger matching/firing may transition the hart to
-Debug Mode. See other cases in RISC-V Debug specification cite:[ExtDbg]. 
-Note that the `resethaltreq` is not applicable since it can be done only
-following a reset and the hart cannot be at privilege mode less than
-`M-mode` with `msdcfg.SDEDBGALW`=1 at de-assertion of reset.
-* Abstract commands and program buffer execution can access state of privilege
-modes less than `M-mode`. In this context, "state" includes all resources
-accessible per the Debug specification cite:[ExtDbg].
-* Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
-* Debugger memory accesses may occur with `S-mode` or `U-mode` privilege per
-software execution in the SD, and as if `aamvirtual` = 1 and `MPP` != `M-mode`
-and `relaxedpriv`=0.
+To enforce the above controls specified by this extension, the following
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
 
-==== `Smsdedbg` interaction with external debug security controls (Informative)
+. The enable control for external debug driven by the external debug module is
+  expected to be established by the platform root-of-trust following RISC-V
+  Security Model recommendations SR_GEN_007 and SR_GEN_012. When the control is
+  in cleared state, the hart should not be able to enter external debug.
+. External debug must be able to transition the hart to Debug Mode and access
+  supervisor domain memory and state. In this context, "state" includes all
+  non-M-mode resources accessible per the Debug specification cite:[ExtDbg].
+. Triggers must match and fire on a hart only when a supervisor domain with this
+  configuration is active on a hart (and not in M-mode).
+. The RDSM must be able to context switch all triggers even when `d-mode` = 1.
+  This functional change allows the RDSM to remain in control of external debug
+  for supervisor domains when it is not under external debug itself.
 
-This section will be moved into the specification for external debug security.
-It is described in this specification as informational.
-
+The following figure is non-normative and is intended to illustrate the use of
+the `msdcfg.sdedbgalw` control - the normative specification will be in the
+specification for external debug security cite:[ExtDbgSec].
 
 [caption="Figure {counter:image}: ", reftext="Figure {image}"]
-[title= "External Debug for Supervisor Domain", id=Smsdedbg_img]
+[title= "External Debug opt-in for Supervisor Domain", id=Smsdedbg_img]
 image::images/Smsdedbg.png[]
-
-
-The `medbgen` is an enable control for external debug for the M-mode driven by
-the debug module and is expected to be established by the RoT (following RISC-V
-Security Model recommendation SR_GEN_007 and SR_GEN_012). When privilege is `M`,
-the `medbgen` gates the `haltreq` from the debug module and if is 0 prevents
-the hart from entering external debug mode.
-
-The following change is proposed to behavior of `M-mode` access to triggers with
-`dmode` = 1. This change allows the RDSM to remain in control of external debug
-for supervisor domains (unless the RDSM is itself under external debug in which
-case this relaxation is a don't care condition).
-
-When `medbgen` is 0 and privilege is `M-mode`:
-
-* M-mode can read and write triggers, *including triggers with `dmode` = 1
-without restrictions*.
-* Abstract commands and requests from an external debugger, that transition the
-hart into a halted state, stay pending while privilege is `M-mode` cite:[ExtDbg].
-* All trigger-matching is prevented from matching or firing, similar to how
-`MIE` or `MTE` would behave. This behavior is specifically as stated in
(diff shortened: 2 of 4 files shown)


rsahita:
cc @AoteJin @ved-rivos

chapter8.adoc
```diff
@@ -6,22 +6,58 @@ The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
...
+When M-mode external debug is enabled, all supervisor domains may also be
+debugged by an external debugger irrespective of the configuration held in
+`msdcfg.SDEDBGALW`.
+
+When M-mode external debug is disabled, whether execution at privilege modes
+less than `M-mode` may be debugged by an external debugger depends on the
+configuration held in `msdcfg.SDEDBGALW`.
+
+When `msdcfg.SDEDBGALW` = 0, external debug is disallowed. Abstract commands
+and halt request from the debug module are suppressed and stay pending.
```
ved-rivos, line 18:
```suggestion
and halt request from the Debug Module are suppressed and stay pending.
```

chapter8.adoc
```diff
@@ -6,22 +6,58 @@ The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
...
+
+* M-mode can read and write triggers, *including triggers with `dmode` = 1
+without restrictions*.
+* Abstract commands and halt request from external debugger stay pending while
+privilege is `M-mode`.
+* All trigger-matching is suppressed (similar to how `MIE` or `MTE` would
+suppress them)
+
+When privilege is less than `M`, the OR of the `MSDCFG.sdedbgalw` and `medbgen`
+gates the `haltreq` from the debug module and the hart will enter external debug
```
ved-rivos, line 59:
```suggestion
gates the `haltreq` from the debug module and the hart will enter Debug
```

chapter8.adoc
```diff
@@ -6,22 +6,58 @@ The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
...
+mode if either is 1.
```
ved-rivos, line 60:
```suggestion
Mode if either is 1.
```

chapter8.adoc
```diff
@@ -30,8 +66,8 @@ security manager using secure memory.
 When `medbgen` is 1, there are no restrictions. When `medbgen` is 0 but
 `MSDCFG.sdedbgalw` is 1, then the external debug mode may be entered but is
 restricted by `M-mode` software to prevent privileged CSR accesses and memory
-accesses by instructions executed in external debug mode cannot use `M` privilege.
-When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the M-mode root domain
-security manager must not configure triggers with `action=1`. Triggers for the
-supervisor domain are expected to be controlled by the root domain security
-manager to prevent any leakage of information.
+accesses by instructions executed in external debug mode cannot use `M-mode`
```
ved-rivos, line 69:
```suggestion
accesses by instructions executed in Debug Mode cannot use `M-mode`
```

chapter8.adoc
```diff
@@ -30,8 +66,8 @@ security manager using secure memory.
...
+privilege. When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the `M-mode`
+root domain security manager must not configure triggers with `action=1`.
+Triggers for the supervisor domain are expected to be controlled by the root
```
ved-rivos, line 72:
```suggestion
Triggers for the supervisor domain must be controlled by the root
```

chapter8.adoc
```diff
@@ -30,8 +66,8 @@ security manager using secure memory.
...
+domain security manager to prevent any leakage of information.
```
ved-rivos, line 73:
```suggestion
domain security manager to prevent any leakage of information across security domains.
```

ved-rivos (approved):
Left some suggestions.

chapter8.adoc
```diff
@@ -30,8 +70,45 @@ security manager using secure memory.
...
+output from the trace encoder module are disabled while the SD is active (the
+trace encoder must behave per `trTeEnable`=0).
+
+When `msdcfg.SDEDBGALW` = 1 then external trace of privilege modes less than
+`M-mode` is allowed for the SD, and programming of `trTeActive` and `trTeEnable`
+to 1 may transition the hart to enable trace for all privilege modes less than
+`M-mode`.
+
+==== `Smsdetrc`: interaction with external trace security controls (Informative)
+
```
ved-rivos, line 99: Why not:
When metrcen is 0, halted is asserted on:
  - Entry to Debug mode.
  - Entry to privilege M.
  - Entry to privilege < M with sdetrcalw == 0.
When metrcen is 0, halted is deasserted on:
  - Entry to privilege < M with sdetrcalw == 0.
When metrcen is 1, halted is asserted on:
  - Entry to Debug mode.
When metrcen is 1, halted is deasserted on:
  - Resume from Debug mode.

rsahita: thats the same as the following 

> When `metrcen` is 0 and privilege is `M-mode`, `M-mode` activity cannot be
> traced and `M-mode` software can enforce that `trTeEnable` is set to the state
> of `msdcfg.SDETRCALW` for the SD it context switches the hart into.

chapter8.adoc
```diff
@@ -30,8 +70,45 @@ security manager using secure memory.
...
+be traced by an external debugger irrespective of the configuration held in
+`msdcfg.SDETRCALW`.
+
+When M-mode external trace is disabled, whether execution at privilege modes
+less than `M-mode` may be traced by an external debugger depends on the
+configuration held in `msdcfg.SDETRCALW`.
+
+When `msdcfg.SDETRCALW` = 0, external trace is disallowed for the SD. All trace
+output from the trace encoder module are disabled while the SD is active (the
+trace encoder must behave per `trTeEnable`=0).
```
bcstrongx, line 91: This should be reviewed by the DTPM SIG.  Current discussion is that trace suppression as a result of METRCEN/SDETRCALW will appear akin to trace suppressed as a result of a hart entering the halted state.  I believe this differs from clearing of trTeEnable by virtue of the indication of why trace stopped.

chapter8.adoc
```diff
@@ -30,8 +70,45 @@ security manager using secure memory.
...
+`M-mode` is allowed for the SD, and programming of `trTeActive` and `trTeEnable`
+to 1 may transition the hart to enable trace for all privilege modes less than
+`M-mode`.
+
+==== `Smsdetrc`: interaction with external trace security controls (Informative)
+
+This section will be moved into the specification for external debug security.
+It is described in this specification as informational.
+
+The `metrcen` is an enable control for external race for the `M-mode` driven by
```
bcstrongx, line 103: race -> trace

chapter8.adoc
```diff
@@ -30,8 +70,45 @@ security manager using secure memory.
...
+an external trace module and is expected to be established by the RoT (following
+RISC-V security model recommendations SR_GEN_007 and 012). When privilege is
+`M-mode`, the `medbgtrc` gates the `trTEActive` control into the hart trace
```
bcstrongx, line 106: medbgrc -> metrcen

chapter8.adoc
rsahita, line 103:
```suggestion
The `metrcen` is an enable control for external trace for the `M-mode` driven by
```

chapter8.adoc
rsahita, line 106:
```suggestion
`M-mode`, the `metrcen` gates the `trTEActive` control into the hart trace
```

chapter8.adoc
```diff
@@ -1,27 +1,67 @@
...
+commands and halt request from the debug module are suppressed and stay pending
+while the SD is active.
+
+When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for the SD, and:
+
+* A halt request may transition the hart to Debug Mode.
+* Abstract commands and program buffer execution can access state of privilege
+modes less than `M-mode`.
+* Read and Write of `Sdtrig` CSRs is allowed.
```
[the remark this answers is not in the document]
rsahita: wanted to be explicit about it - are you saying, if we do mention it, we should say:

> Read and Write of debug-related CSRs is allowed

chapter8.adoc
rsahita: I see - but what is the control that suppresses the trace from the trace encoder - per the trace control interface that seems to be trTeEnable?

chapter8.adoc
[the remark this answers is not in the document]
ved-rivos: I think we   should add "Sdext" CSRs also.

rsahita, line 30:
```suggestion
* Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
```

chapter8.adoc
```diff
@@ -1,27 +1,67 @@
...
 [[Smsdedbg]]
-== `Smsdedbg`: Supervisor Domain External Debug
+== Supervisor Domain External Trace and Debug
 
-The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
-current scheduled SD is allowed to be external-debugged cite:[ExtDbg]. This bit
-is context switched (along with rest of the `msdcfg`) per SD.
+The <<MSDCFG>> CSR contains the `sdedbgalw` and `sdetrcalw` bits that control
+whether the current scheduled SD is allowed to be external-debugged or to be
+external-traced (independently). These control bits are context switched (as
```
ved-rivos, line 7: (independently) -> respectively. 
The last sentence can be removed as it is implied by "current scheduled SD"

chapter8.adoc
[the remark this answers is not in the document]
rsahita: ack

chapter8.adoc
```diff
@@ -1,27 +1,67 @@
...
+part of the `msdcfg`) per SD.
```
rsahita, line 8:
```suggestion
external-traced respectively.
```

chapter8.adoc
```diff
@@ -30,8 +70,45 @@ security manager using secure memory.
...
-supervisor domain are expected to be controlled by the root domain security
-manager to prevent any leakage of information.
+accesses by instructions executed in Debug Mode cannot use `M-mode`
+privilege. When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the `M-mode`
+root domain security manager must not configure triggers with `action=1`.
+Triggers for the supervisor domain must be controlled by the root
+domain security manager to prevent any leakage of information across security domains.
+
+=== `Smsdetrc`: External Trace allowed control for Supervisor Domain
+
```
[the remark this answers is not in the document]
rsahita: the "halted" signal being reused in both debug and trace spec's is confusing.
Now that I understand you are referring to the trace optional side-band "halted" going from the hart to the trace encoder (per the etrace spec) - i think this makes sense. Also will mention that this optional side-band signal will be required when supervisor domains are implemented.

chapter8.adoc
rsahita: ack

chapter8.adoc
```diff
@@ -102,12 +103,27 @@ It is described in this specification as informational.
...
-`M-mode`, the `metrcen` gates the `trTEActive` control into the hart trace
-encoder and if 0, prevents the hart from generating any trace output.
-
-When `metrcen` is 0 and privilege is `M-mode`, `M-mode` activity cannot be
-traced and `M-mode` software can enforce that `trTeEnable` is set to the state
-of `msdcfg.SDETRCALW` for the SD it context switches the hart into.
-
-When `metrcen` is 1, there are no restrictions and all privilege modes
-(inclusive of `M-mode` and any SDs managed by the RDSM) are traceable.
+`M-mode`, the `metrcen` gates the `halted` signal (from the hart) into the hart
```
ved-rivos, line 106:
```suggestion
`M-mode`, the `metrcen` determines the state of the `halted` side band signal sent by the hart to the
```

chapter8.adoc
```diff
@@ -73,7 +73,8 @@ accesses by instructions executed in Debug Mode cannot use `M-mode`
 privilege. When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the `M-mode`
 root domain security manager must not configure triggers with `action=1`.
 Triggers for the supervisor domain must be controlled by the root
-domain security manager to prevent any leakage of information across security domains.
+domain security manager to prevent any leakage of information across security
```
ved-rivos, line 76:
```suggestion
domain security manager to prevent leakage of information across security
```

ved-rivos (approved):
Left a couple of comments. Otherwise LGTM.

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+`dmode` = 1. This change allows the RDSM to remain in control of external debug
+for supervisor domains (unless the RDSM is itself under external debug).
+
+When `medbgen` is 0 and privilege is `M-mode`:
+
+* M-mode can read and write triggers, *including triggers with `dmode` = 1
+without restrictions*.
+* Abstract commands and halt request from external debugger stay pending while
+privilege is `M-mode`.
+* All trigger-matching is suppressed (similar to how `MIE` or `MTE` would
```
AoteJin, line 58: It is not comparable. The incoming interrupt blocked by MIE will be pending and fire when it is allowed (e.g. entering a less privileged mode). But a matching trigger in M mode should NOT be pending and enter Debug Mode when it enters SD. Neither should icout trigger increment if it hit interrupt in M mode.
The word "suppress" is ambiguous. It must be clearly stated whether it pends or is dropped.

chapter8.adoc
```diff
@@ -30,8 +69,64 @@ security manager using secure memory.
 When `medbgen` is 1, there are no restrictions. When `medbgen` is 0 but
 `MSDCFG.sdedbgalw` is 1, then the external debug mode may be entered but is
 restricted by `M-mode` software to prevent privileged CSR accesses and memory
```
AoteJin, line 71: "restricted by `M-mode` software" is a confusing statement. The extension is a HW/SW co-work solution, not pure software thing.

chapter8.adoc
```diff
@@ -30,8 +69,64 @@ security manager using secure memory.
...
-accesses by instructions executed in external debug mode cannot use `M` privilege.
-When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the M-mode root domain
-security manager must not configure triggers with `action=1`. Triggers for the
-supervisor domain are expected to be controlled by the root domain security
-manager to prevent any leakage of information.
+accesses by instructions executed in Debug Mode cannot use `M-mode`
```
AoteJin, line 72: Not only instructions in program buffer  but also abstract commands

chapter8.adoc
```diff
@@ -30,8 +69,64 @@ security manager using secure memory.
...
+privilege. When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the `M-mode`
+root domain security manager must not configure triggers with `action=1`.
```
AoteJin, line 74: Even if the triggers are configure with `action=1`, it does yield any side effect, because it just cannot enter Debug Mode. The restriction her  seems redundant.

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
 Security Model recommendation SR_GEN_007 and SR_GEN_012). When privilege is `M`,
 the `medbgen` gates the `haltreq` from the debug module and if is 0 prevents
-the hart from entering external debug mode. When privilege is less than `M`, the
-OR of the `MSDCFG.sdedbgalw` and `medbgen` gates the `haltreq` from the debug
-module and the hart will enter external debug mode if either is 1.
+the hart from entering external debug mode.
+
+The following change is proposed to behavior of `M-mode` access to triggers with
+`dmode` = 1. This change allows the RDSM to remain in control of external debug
+for supervisor domains (unless the RDSM is itself under external debug).
```
AoteJin, line 50: What does "unless" implies? Do you mean that when medbgen=1, the M-mode SW still cannot write triggers if dmode=1?
If so, the behavior needs to be clearly outlined.

AoteJin (changes requested):

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
 [[chapter8]]
 [[Smsdedbg]]
-== `Smsdedbg`: Supervisor Domain External Debug
+== Supervisor Domain External Trace and Debug
```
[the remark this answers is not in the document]
rsahita: defined

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+When M-mode external debug is enabled, all supervisor domains may also be
+debugged by an external debugger irrespective of the configuration held in
+`msdcfg.SDEDBGALW`.
+
+When M-mode external debug is disabled, whether execution at privilege modes
+less than `M-mode` may be debugged by an external debugger depends on the
+configuration held in `msdcfg.SDEDBGALW`.
+
+When `msdcfg.SDEDBGALW` = 0, external debug is disallowed for the SD. Abstract
+commands and halt request from the debug module are suppressed and stay pending
```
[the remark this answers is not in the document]
rsahita: added a note to serve as a catch all for other debug operations (currently defined or may be defined in the future) must be kept pending while the SD that has opt'ed out of external debug is active on a hart.

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+while the SD is active.
+
+When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for the SD, and:
+
+* A halt request may transition the hart to Debug Mode.
+* Abstract commands and program buffer execution can access state of privilege
```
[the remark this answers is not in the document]
rsahita: expanded to clarify that "state" includes all resources accessible per the Debug spec (added a citation as well).

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+modes less than `M-mode`.
+* Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
```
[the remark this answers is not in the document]
rsahita: so the statement is qualified by the title of the header of the bullet points - specifically - "When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than `M-mode` is allowed for the SD on a per-hart basis, and:" -> Read and Write of `Sdtrig` and `Sdext` CSRs are allowed . 
Once the SD opts in to external debug, all rules specified in debug spec for CSR accesses by the SD or by M on behalf of the SD apply so we don't have to restate those IMO.

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
 
-This section will be moved into the non-ISA specification for external debug
-security. It is described here as informational.
+This section will be moved into the specification for external debug security.
+It is described in this specification as informational.
 
 The `medbgen` is an enable control for external debug for the M-mode driven by
 the debug module and is expected to be established by the RoT (following RISC-V
 Security Model recommendation SR_GEN_007 and SR_GEN_012). When privilege is `M`,
 the `medbgen` gates the `haltreq` from the debug module and if is 0 prevents
```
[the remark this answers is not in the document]
rsahita: I presume you are specifying the additional rules in the secure external debug spec ISA - in this spec I am only specifying the msdcfg control bits and how it figures into the gating of haltreq.
I will add a reference to the ext secure debug spec (once it is available).

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+* Abstract commands and halt request from external debugger stay pending while
+privilege is `M-mode`.
+* All trigger-matching is suppressed (similar to how `MIE` or `MTE` would
+suppress them)
+
+When privilege is less than `M`, the OR of the `MSDCFG.sdedbgalw` and `medbgen`
+gates the `haltreq` from the debug module and the hart will enter Debug
+Mode if either is 1.
 
 The configuration for `MSDCFG.sdedbgalw` may be obtained from the manifest/
```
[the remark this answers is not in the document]
rsahita: right - i put it into a note.

chapter8.adoc
```diff
@@ -30,8 +69,64 @@ security manager using secure memory.
...
+privilege. When `medbgen` is 0 and `MSDCFG.sdedbgalw` is also 0, the `M-mode`
+root domain security manager must not configure triggers with `action=1`.
+Triggers for the supervisor domain must be controlled by the root
+domain security manager to prevent leakage of information across security
+domains.
+
+=== `Smsdetrc`: External Trace allowed control for Supervisor Domain
+
+When M-mode external trace is enabled, all supervisor domains activity may also
+be traced by an external debugger irrespective of the configuration held in
```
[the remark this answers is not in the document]
rsahita: fixed

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
 [[Smsdedbg]]
-== `Smsdedbg`: Supervisor Domain External Debug
+== Supervisor Domain External Trace and Debug
 
-The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
-current scheduled SD is allowed to be external-debugged cite:[ExtDbg]. This bit
-is context switched (along with rest of the `msdcfg`) per SD.
+The <<MSDCFG>> CSR contains the `sdedbgalw` and `sdetrcalw` bits that control
+whether the current scheduled SD is allowed to be external-debugged or to be
+external-traced respectively.
```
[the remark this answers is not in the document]
rsahita: yes this extension related only to external debug and external trace interactions with supervisor domains - added a note in the introduction of this section.

chapter8.adoc
rsahita: clarified that it is prevented from matching.

chapter8.adoc
rsahita: agree - added clarification that M-mode software controls the policy via the msdcfg control bit sdedbalw which gates external debug.

chapter8.adoc
rsahita: ack.

chapter8.adoc
rsahita: removed this restriction and clarified with the enforcement that M-mode software should ensure that filter mcontrol.m bit should be clear.

chapter8.adoc
rsahita: clarified that this feature relaxation takes effect only if medbgen is deasserted, o/w if the RDSM is itself under debug due to medbgen being asserted then this change is don't care.

chapter8.adoc
```diff
@@ -1,37 +1,146 @@
...
+==== `Smsdetrc`: interaction with external trace security controls (Informative)
+
+This section will be moved into the specification for external debug security.
+It is described in this specification as informational.
+
+The `metrcen` is an enable control for external trace for the `M-mode` driven by
+an external trace module and is expected to be established by the RoT (following
+RISC-V security model recommendations SR_GEN_007 and 012). When privilege is
+`M-mode`, the `metrcen` determines the state of the `halted` side band signal sent by the hart to the
+trace encoder. Per the Efficient trace specification cite:[Etrc], this side-band
```
rsahita, line 121:
```suggestion
trace encoder. Per the Efficient trace specification cite:[ETrc], this side-band
```

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: It introduce security holes. If you look at sdtrig CSR tdata1. It could set trigger for machine mode (e.g. set trigger to raise exception in M-mode when it is not yet context switched). In your statement, it just expose the privilege escalation attack vector. 
BTW, what do you mean by "M on behalf of the SD"?

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: could you specifically point out where it is defined? I could not find any.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: Not all debug operation pends. For example, a trigger set for M-mode to enter Debug Mode  cannot pend if `medbgen` is 0, because it might block other triggers to take effect which is configured for submachine mode.

chapter8.adoc
```diff
@@ -1,37 +1,146 @@
...
+a hart.
+
+When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for the SD on a per-hart basis, and:
+
+* A valid request e.g. `haltreq` may transition the hart to Debug Mode. See
+other cases in RISC-V Debug specification cite:[ExtDbg].
+* Abstract commands and program buffer execution can access state of privilege
+modes less than `M-mode`. In this context, "state" includes all resources
+accessible per the Debug specification cite:[ExtDbg].
```
AoteJin, line 36: It is contradictory, debug spec states the it have "God" mode privilege and could access all resources, not constrained by privilege modes.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: It is very confusing. Because in other paragraph (above and blow), the trigger, abstract commands are mentioned. I don't think user can get the point "in this spec I am only specifying the msdcfg control bits and how it figures into the gating of haltreq."

AoteJin (changes requested):

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: Your point extends the scope of the current trace spec very much. The trace control now is defined could either be accessed "internally" and "externally". They will be both affected by "halted" signal, thus your point "this extension related only to external debug and external trace interactions with supervisor domains" cannot be interpreted correctly if not otherwise explained. 
The idea of isolating "external" and "internal" must involve sufficient DTPM discussion and cannot easily made.

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+
+When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for the SD, and:
+
+* A halt request may transition the hart to Debug Mode.
+* Abstract commands and program buffer execution can access state of privilege
+modes less than `M-mode`.
+* Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
+* Debugger memory accesses occur with either `S-mode` or `U-mode` privilege (as
+if `aamvirtual` = 1 and `MPP` != `M-mode`).
```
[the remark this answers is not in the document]
AoteJin: The problem brought up here is not yet addressed. It will break debug use cases.

chapter8.adoc
```diff
@@ -30,8 +69,64 @@ security manager using secure memory.
 When `medbgen` is 1, there are no restrictions. When `medbgen` is 0 but
```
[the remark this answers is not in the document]
AoteJin: Please pay attention to this comment

chapter8.adoc
rsahita: sorry I don't understand what you are referring to as contradictory - can you clarify?
we are stating when M-mode external debug is disabled, whether execution at privilege modes less than `M-mode` may be debugged by an external debugger depends on the configuration held in `msdcfg.SDEDBGALW` -  so only when additionally `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than `M-mode` is allowed for the SD on a per-hart basis, and within that scope all state of the SD can be accessed.

chapter8.adoc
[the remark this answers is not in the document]
rsahita: "M on behalf of the SD" -> means RDSM enforcing any rules since it is in the TCB of the SDs.
an e.g. for this is -  we just added a note that for any triggers a SD sets up, RDSM must ensure the m bit is cleared so that prevents the scenario you mentioned.

can you clarify what security holes you see?

chapter8.adoc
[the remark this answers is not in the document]
rsahita: I can add a note to that effect in the intro such that the reader is aware that they should read the "secure external debug spec" as well to get the full picture.

chapter8.adoc
AoteJin: "clarified with the enforcement that M-mode software should ensure that filter mcontrol.m bit should be clear." This restriction should not be made. It will de-feature self-hosted debug for machine mode

AoteJin (changes requested):

chapter8.adoc
AoteJin: I mean ""state" includes all resources accessible per the Debug specification", contradicts the "access state of privilege modes less than `M-mode`".
People get confused when read smmtt and debug spec

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: Sure, of course. What I mentioned above is exactly an example of hole and it cannot be prevented by RDSM, because the external debugger set it without involvement of RDSM.

chapter8.adoc
rsahita: will it help to restate as: 
"* Abstract commands and program buffer execution can access state of the SD that opted in to allow external debug. Here state includes system resources of privilege modes less than `M-mode` accessible per the Debug specification cite:[ExtDbg] ." 
- hope this clearly qualifies the statement.

chapter8.adoc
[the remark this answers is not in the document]
rsahita: M-mode needs to protect itself in this case. which is we have the new proposed behavior change (that I hope you will pull into the ext secure debug spec), specifically:

When `medbgen` is 0 and privilege is `M-mode`:
* M-mode can read and write triggers, *including triggers with `dmode` = 1 without restrictions*.
* Abstract commands and requests from an external debugger, that transition the hart into a halted state, stay pending while privilege is `M-mode` cite:[ExtDbg].
* All trigger-matching is prevented from matching or firing, similar to how `MIE` or `MTE` would behave.

chapter8.adoc
AoteJin: Actually the statement go against the debug specification. Instead of "per the Debug specification", it should be per risc-v privileged spec.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: It helps little. The right direction is if the feature/behavior is mentioned in two spec, it should be described thoroughly in both.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: `MIE` and `MTE` solution is to tackle re-entrance problem (with action=0), while the issue I mentioned in my previous post is about the trigger traps into M-mode. Re-entrance issue only happens after it. 
More importantly, the statement "similar to how MIE or MTE would behave." is misleading. I thought it means:
- similar to action=0, the triggers with action=1 are prevented from firing/matching to enter Debug Mode

But based on your comment, seems it is supposed to convey different idea:

- triggers with action=1 are prevented from enter debug mode in M-mode
- triggers with action=0 are prevented from raising exception in M-mode

The second bullet is brand new idea and never bought up. I think we need involve more people discuss about the scope it affected. 

Above all, "similar to how MIE or MTE would behave." has to be taken out. It gives misleading idea.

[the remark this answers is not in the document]
AoteJin: "All trigger-matching" - could implies address/data/exception/interrupt matching (which is arguably intuitive for readers). It is not a good point to keep ambiguous statement based on personal understanding. 

It is a brand new in the scope of external debug security.
If you quote the whole context, it could be easily identified that the Debug spec introduce them to solve re-entrance issue, which prevents exception from same privilege (e.g. M-mode->M-mode).
But in scope of external debug, the implication is to prevent the exception from lower privilege (e.g. S-mode -> M-mode). Both the hardware implementation and software use case are different.

chapter8.adoc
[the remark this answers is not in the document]
ved-rivos: The key word is "as if". The "Access Memory" command is specified in section 3.7.1.3. This command lets debugger perform memory access. The arg0 of this command has aamvirtual which when set to 1 indicates the address is a virtual address - and these are translated the same way as they would be from M-mode, with MPRV set. The abstract commands can only execute at U or S mode when medbgen is 0 and they will always execute as-if aamvirtual = 1 and MPP !=M-mode. Please clarify what debug use cases get affected.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: If I didn't make the point clear enough, we could probably discuss it further in tomorrow's external debug TG meeting? 

[https://zoom.us/j/99483621157?pwd=cXYvT0I4QXBMdEZlVkhjbnRtUEdrUT09](https://www.google.com/url?q=https://zoom.us/j/99483621157?pwd%3DcXYvT0I4QXBMdEZlVkhjbnRtUEdrUT09&sa=D&source=calendar&usd=2&usg=AOvVaw3-CCTBCDyWhEIIJEWGPueR)

chapter8.adoc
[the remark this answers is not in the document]
ved-rivos: When medbgen is 1 then there are no restrictions on the type of debug. For example, the debugger can gain run control and single step M-mode and change its register state. This does not mean that the debugger can do what it could not otherwise do such as writing to a ROM, or  changing a locked PMP.

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+
+The following change is proposed to behavior of `M-mode` access to triggers with
+`dmode` = 1. This change allows the RDSM to remain in control of external debug
+for supervisor domains (unless the RDSM is itself under external debug).
+
+When `medbgen` is 0 and privilege is `M-mode`:
+
+* M-mode can read and write triggers, *including triggers with `dmode` = 1
+without restrictions*.
+* Abstract commands and halt request from external debugger stay pending while
```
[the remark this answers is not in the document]
ved-rivos: Only two aspects matter here - issuing abstract commands and issuing a halt request. The ability of the debug module to assert and deassert reset has to be controlled in the debug module. Please clarify more what "locked PMP access" means and what is included in "etc.".

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: Exactly, your comments has to explicitly address in spec, because the Debug spec defines the behavior to bypass locked PMP. Thus, obviously "no restrictions" cannot be said.

chapter8.adoc
ved-rivos: I think the "all resources accessible" is the right term. There are 3 abstract commands - access register, quick access, and access memory. The access register command is used to access abstract registers (table 4) - CSRs, GPRs, and floating point registers. The quick access is used to execute the program buffer. Access memory is used to access memory. Program buffer execution can be used to execute instructions. The CSR, GPRs, and FP registers must be the state of the SD i.e. CSRs cannot be M-mode or D-mode CSRs. The instructions executed from program buffer cannot access any M-mode CSRs.

chapter8.adoc
```diff
@@ -1,27 +1,66 @@
...
+configuration held in `msdcfg.SDEDBGALW`.
+
+When `msdcfg.SDEDBGALW` = 0, external debug is disallowed for the SD. Abstract
+commands and halt request from the debug module are suppressed and stay pending
+while the SD is active.
+
+When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for the SD, and:
+
+* A halt request may transition the hart to Debug Mode.
```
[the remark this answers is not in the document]
ved-rivos: This bullet can be expanded to say "A halt request or a trigger matching/firing.."
The resethaltreq is not applicable - reset halt can be done only following a reset and one cannot be privilege mode less than M with msdcfg.SDEDBGALW=1 at deassertion of reset.

chapter8.adoc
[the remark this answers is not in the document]
ved-rivos: I suggest using the precise words from section 5.4 and section 5.7.6, repeated above, rather than loose re-interpretations like  "which prevents exception from same privilege".

chapter8.adoc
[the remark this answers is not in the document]
ved-rivos: The debug specification allows for bypassing a locked PMP and it should be when M is under external debug. There is no  need to restrict anything that can be done in debug mode when external debug is enabled. Bypassing PMP is no worse than executing instructions of debuggers choice from the program buffer and having full run control on M-mode.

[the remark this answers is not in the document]
AoteJin: There is no definition in debug spec to support the argument "The debug specification allows for bypassing a locked PMP and it should be when M is under external debug. "

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: Before driving this topic off-track, I would strongly suggest let's discuss it on TG meeting . Honestly, I don't think we are talking about the same thing.

chapter8.adoc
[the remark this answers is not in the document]
ved-rivos: > There is no definition in debug spec to support the argument "The debug specification allows for bypassing a locked PMP and it should be when M is under external debug. "

Anything and everything that the debug specification allows to occur for M-mode external debug should be allowed when M-mode is under external debug. This includes single stepping M-mode to and skipping over the locking of the PMPs themself.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: The statement says "S-mode or U-mode", but basically which one? According to value in MPP or what else? The spec needs clear definition on access privilege. Otherwise, it leaves impression that use cases are affected.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: >There is no definition in debug spec to support the argument "The debug specification allows for bypassing a locked PMP and it should be when M is under external debug. "
Sorry for confusion, what I mean is debug spec doesn't have definition to support  argument "....when M is under external debug", that is when sub-M mode is allowed to debug, relaxedpriv could also be set. 

That's why no restrictions" cannot be said. It has to restrict relaxedpriv.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: This point must be clearly defined in main content. But currently not a clue.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: "one cannot be privilege mode less than M with msdcfg.SDEDBGALW=1 at deassertion of reset." This point ought to be clearly included in main text.

chapter8.adoc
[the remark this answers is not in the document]
AoteJin: > Only two aspects matter here - issuing abstract commands and issuing a halt request. The ability of the debug module to assert and deassert reset has to be controlled in the debug module. Please clarify more what "locked PMP access" means and what is included in "etc.".

Triggers, SBA, and probably something else.

chapter8.adoc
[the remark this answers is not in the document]
ved-rivos: That is unrelated to this comment thread which is about medbgen=1. 
I agree that the "as if `aamvirtual` = 1 and `MPP` != `M-mode`" should be expanded "as if `aamvirtual` = 1 and `MPP`!=`M-mode` and `relaxedpriv`=0"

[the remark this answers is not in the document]
AoteJin: Sorry for the deviation. 
Let me rephrase it:
even for the mdbgen=1, relaxedpriv has to be restricted.

>"Bypassing PMP is no worse than executing instructions of debuggers choice from the program buffer and having full run control on M-mode."

1. M-mode cannot bypass locked PMP
2. relaxedpriv affect program buffer as well

Overall, with `relaxedpriv` uncontrained give the external debugger the capability even beyond the M-mode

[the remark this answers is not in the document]
ved-rivos: I don't see motivation for doing that when medbgen is 1. When medbgen is 1 all M-mode resources are fair game. medbgen can be used to prevent the PMP from being locked by M-mode in the first place.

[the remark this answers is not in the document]
AoteJin: There is lifecycle requirement to mandate it. Even for M-mode, there are potentially multiple actors. Locked PMP can never be bypassed.
logical_sec_model_fig1

chapter8.adoc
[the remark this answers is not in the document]
rsahita: Again, the main spec should be the ext. secure debug spec - we cannot repeat information in 2 specs - that will lead to inconsistency and bugs - the goal of this supervisor domains spec chapter is to describe the additional requirements i.e. 2 things: 1. the controls specified in msdcfg for sdedbgalw and sdetrcalw and how they feed into/gate medbgen  2. the behavior change in m-mode to allow it to access debug CSRs when d-mode=1 so that when it is not under external debug it can manage ext debug of a SD that opts in. 

can we agree on this partitioning of the spec first? cc @joxie

chapter8.adoc
```diff
@@ -1,37 +1,146 @@
...
+
+When M-mode external trace is disabled, whether execution at privilege modes
+less than `M-mode` may be traced by an external trace tool depends on the
+configuration held in `msdcfg.SDETRCALW`.
+
+When `msdcfg.SDETRCALW` = 0, external trace is disallowed for the SD. All trace
+output from the trace encoder module are disabled while the SD is active on a
+per-hart basis.
+
+When `msdcfg.SDEDBGALW` = 1 then external trace of privilege modes less than
```
AoteJin, line 109: msdcfg.SDEDBGALW -> msdcfg.SDETRCALW

chapter8.adoc
gokhankaplayan: In addition to CSRs, memory access should be protected by PMP, MTT. E.g. a SD should not be able to access other SD's memory or M-mode memory (if memory region is not shared). 
In summary, referring to ExtDbg spec seems the correct way to me.

chapter8.adoc
```diff
@@ -1,37 +1,146 @@
...
+commands and halt request from the debug module are suppressed and stay pending
+while the SD is active, on a per-hart basis. Other debug operations (current or
+defined in the future) must similarly be kept pending while the SD is active on
+a hart.
+
+When `msdcfg.SDEDBGALW` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for the SD on a per-hart basis, and:
+
+* A valid request e.g. `haltreq` may transition the hart to Debug Mode. See
+other cases in RISC-V Debug specification cite:[ExtDbg].
```
rsahita, line 33:
```suggestion
* A valid request (e.g. `halt`) or a trigger matching/firing may transition the hart to
Debug Mode. See other cases in RISC-V Debug specification cite:[ExtDbg]. 
Note that the `resethaltreq` is not applicable since it can be done only
following a reset and the hart cannot be at privilege mode less than
`M-mode` with `msdcfg.SDEDBGALW`=1 at de-assertion of reset.
```

chapter8.adoc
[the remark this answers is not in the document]
rsahita: fixed.

chapter8.adoc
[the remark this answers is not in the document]
rsahita: please clarify what is unclear in the statement in the text:
Debugger memory accesses occur with either `S-mode` or `U-mode` privilege (as if `aamvirtual` = 1 and `MPP` != `M-mode`).

chapter8.adoc
```diff
@@ -1,37 +1,149 @@
...
+Debug Mode. See other cases in RISC-V Debug specification cite:[ExtDbg]. 
+Note that the `resethaltreq` is not applicable since it can be done only
+following a reset and the hart cannot be at privilege mode less than
+`M-mode` with `msdcfg.SDEDBGALW`=1 at de-assertion of reset.
+* Abstract commands and program buffer execution can access state of privilege
+modes less than `M-mode`. In this context, "state" includes all resources
+accessible per the Debug specification cite:[ExtDbg].
+* Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
+* Debugger memory accesses may occur with `S-mode` or `U-mode` privilege per
+software execution in the SD, and as if `aamvirtual` = 1 and `MPP` != `M-mode`.
```
rsahita, line 42:
```suggestion
* Debugger memory accesses may occur with `S-mode` or `U-mode` privilege per
software execution in the SD, and as if `aamvirtual` = 1 and `MPP` != `M-mode`
and `relaxedpriv`=0.
```

chapter8.adoc
```diff
@@ -1,37 +1,150 @@
...
 [[Smsdedbg]]
-== `Smsdedbg`: Supervisor Domain External Debug
+== Supervisor Domain External Trace and Debug
 
-The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
-current scheduled SD is allowed to be external-debugged cite:[ExtDbg]. This bit
-is context switched (along with rest of the `msdcfg`) per SD.
+The <<MSDCFG>> CSR contains the `sdedbgalw` and `sdetrcalw` bits that control
+if the current scheduled SD is allowed to be external-debugged cite:[ExtDbg]
+or to be external-traced, respectively. External trace refers to tracing
```
rsahita, line 7:
```suggestion
if the current scheduled SD is allowed to be external-debugged or to be
external-traced cite:[ExtDbg], respectively. External trace refers to tracing
```

chapter8.adoc
```diff
@@ -1,37 +1,146 @@
 [[chapter8]]
 [[Smsdedbg]]
-== `Smsdedbg`: Supervisor Domain External Debug
+== Supervisor Domain External Trace and Debug
 
-The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
-current scheduled SD is allowed to be external-debugged cite:[ExtDbg]. This bit
-is context switched (along with rest of the `msdcfg`) per SD.
+The <<MSDCFG>> CSR contains the `sdedbgalw` and `sdetrcalw` bits that control
+if the current scheduled SD is allowed to be external-debugged cite:[ExtDbg]
```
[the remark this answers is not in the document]
rsahita: ack

chapter8.adoc
```diff
@@ -1,37 +1,146 @@
...
+or to be external-traced, respectively. External trace refers to tracing
+software running on a separate target RISC-V platform, via a trace control/data
+transport that provides external access to the trace encoder controls
+cite:[ETrc]. This extension only affects debug or trace orchestrated by an
```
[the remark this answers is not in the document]
rsahita: reason i did not do that once since we don't refer to anything from the TCI in this spec per se.

chapter8.adoc
```diff
@@ -1,37 +1,150 @@
...
+case this relaxation is a don't care condition).
 
+When `medbgen` is 0 and privilege is `M-mode`:
+
+* M-mode can read and write triggers, *including triggers with `dmode` = 1
+without restrictions*.
+* Abstract commands and requests from an external debugger, that transition the
+hart into a halted state, stay pending while privilege is `M-mode` cite:[ExtDbg].
+* All trigger-matching is prevented from matching or firing, similar to how
+`MIE` or `MTE` would behave.
```
rsahita, line 72:
```suggestion
* All trigger-matching is prevented from matching or firing, similar to how
`MIE` or `MTE` would behave. This behavior is specifically as stated in
section 5.4 of cite:[ExtDbg]: While `MIE` in `mstatus` is 0, the hardware
prevents triggers with `action`=0 from matching or firing while in `M-mode` .
Also from section 5.7.6 of cite:[ExtDbg], M-mode trigger enable `mte`=0
(disabled) specifies that triggers with `action`=0 do not match/fire
while the hart is in M-mode.
```

chapter8.adoc
```diff
@@ -1,37 +1,155 @@
...
-The <<MSDCFG>> CSR contains the `sdedbgalw` bit that controls whether the
-current scheduled SD is allowed to be external-debugged cite:[ExtDbg]. This bit
-is context switched (along with rest of the `msdcfg`) per SD.
+The <<MSDCFG>> CSR contains the `sdedbgalw` and `sdetrcalw` bits that control
+if the current scheduled SD is allowed to be external-debugged or to be
+external-traced cite:[ExtDbg], respectively. External trace refers to tracing
+software running on a separate target RISC-V platform, via a trace control/data
+transport that provides external access to the trace encoder controls
+cite:[ETrc]. This extension only affects debug or trace orchestrated by an
+external actor. Self-hosted debug (and trace) are managed by the RDSM.
```
rsahita, line 12:
```suggestion
This sub-specification refers to the Debug specification cite:[ExtDbg],
Trace specification cite:[ETrc], and the External debug security extension
cite:[ExtDbgSec] for the description and behavior of controls outside
the scope of this specification, but which interact with controls
specified in this specification when supervisor domains are used.
```

chapter8.adoc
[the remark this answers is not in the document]
rsahita: I added normative references to section 5.4 and 5.7.6 of Debug spec.

example.bib
```diff
@@ -24,6 +24,11 @@ @electronic{ExtDbg
   url = {https://github.com/riscv/riscv-debug-spec}
 }
 
+@electronic{ETrc,
+  title = {RISC-V Efficient Trace for RISC-V, v2.0.2, March 5, 2024},
+  url = {https://github.com/riscv-non-isa/riscv-trace-spec/releases/download/v2.0.2/riscv-trace-spec-asciidoc.pdf}
+}
+
```
rsahita, line 31:
```suggestion

@electronic{ExtDbgSec,
  title = {RISC-V External Debug Security, v0.0, March 26, 2024},
  url = {https://github.com/riscv-non-isa/riscv-external-debug-security/blob/main/external-debug-security.pdf}
}

```

chapter8.adoc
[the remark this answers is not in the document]
rsahita: Added a note in the intro and a citation for https://github.com/riscv-non-isa/riscv-external-debug-security/blob/main/external-debug-security.pdf

rsahita:
@gokhankaplayan made updates per your comments - PTAL

chapter8.adoc
rsahita: not really - this is a software policy completely under m-mode control - don't see how it de-features self-hosted debug at all.

chapter8.adoc
```diff
@@ -1,37 +1,160 @@
...
+`M-mode` with `msdcfg.SDEDBGALW`=1 at de-assertion of reset.
+* Abstract commands and program buffer execution can access state of privilege
+modes less than `M-mode`. In this context, "state" includes all resources
+accessible per the Debug specification cite:[ExtDbg].
+* Read and Write of `Sdtrig` and `Sdext` CSRs are allowed.
+* Debugger memory accesses may occur with `S-mode` or `U-mode` privilege per
+software execution in the SD, and as if `aamvirtual` = 1 and `MPP` != `M-mode`
+and `relaxedpriv`=0.
+
+==== `Smsdedbg` interaction with external debug security controls (Informative)
```
rsahita, line 50:
```suggestion
==== `Smsdedbg` interaction with external debug security controls (Informative)

This section will be moved into the specification for external debug security.
It is described in this specification as informational.

```

chapter8.adoc
```diff
@@ -1,37 +1,160 @@
...
 
 [caption="Figure {counter:image}: ", reftext="Figure {image}"]
 [title= "External Debug for Supervisor Domain", id=Smsdedbg_img]
 image::images/Smsdedbg.png[]
 
-This section will be moved into the non-ISA specification for external debug
-security. It is described here as informational.
+This section will be moved into the specification for external debug security.
+It is described in this specification as informational.
```
rsahita, line 57:
```suggestion
```

chapter3.adoc
```diff
@@ -134,12 +134,21 @@ after writing `mttp`.
...
-the external debug allowed state for a supervisor domain is controlled via the
-`msdcfg.SDEDBALW` bit. Details of `Smsdia` and `Smsdedbg` are described in
-their respective sections in this specification.
+capabilities of the supervisor domain when associated with a hart. +
+The following extensions use the `msdcfg` register to specify additional
+configuration for supervisor domains:
+
+. `Smsdia` uses `msdcfg.SDICN` to specify the active configuration for
+  the supervisor domain interrupt controller associated with the hart.
+. External-debug-allowed state for a supervisor domain is managed via the
```
ved-rivos, line 143: The next two bullets would want to use similar text as first and last bullet in specifying the extension names and field.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
 [[chapter8]]
 [[Smsdedbg]]
-== `Smsdedbg`: Supervisor Domain External Debug
+== Supervisor Domain External Trace and Debug
```
ved-rivos, line 3: why not make two sections - one per extension

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+configuration held in `msdcfg.sdedbgalw`, as described below:
+
+When `msdcfg.sdedbgalw` = 0, external debug is disallowed for the supervisor
+domain. Abstract commands and halt request from the debug module are suppressed
+and stay pending while the supervisor domain is active, on a per-hart basis.
+Other debug operations (current or defined in the future) must similarly be kept
+pending while the supervisor domain is active on a hart. Triggers for the
+supervisor domain must be controlled by the RDSM to prevent any leakage of
+information across security domains.
+
```
ved-rivos, line 50: This para can be replaced with:
When `msdcfg.sdedbgalw` is 0:
- Access by external debuggers to the memory and/or state of the supervisor domain is disallowed.
- Entry to Debug Mode from a supervisor domain is disallowed.

(Do not need to say anything special about triggers).

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+When `msdcfg.sdedbgalw` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for such a supervisor domain on a per-hart basis. +
+
```
ved-rivos, line 53: When msdcfg.sdedbgalw is 1:
1. Access by external debuggers to the memory and/or state of the supervisor domain is allowed.
2. Entry to Debug Mode from a supervisor domain is allowed.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
+
+. The enable control for external debug driven by the external debug module is
+  expected to be established by the platform root-of-trust following RISC-V
+  Security Model recommendations SR_GEN_007 and SR_GEN_012. When the control is
+  in cleared state, the hart should not be able to enter external debug.
+. External debug must be able to transition the hart to Debug Mode and access
+  supervisor domain memory and state. In this context, "state" includes all
```
ved-rivos, line 64: Transition to debug mode is not a pre-requisite for accessing memory or state. Suggest removing this bullet since this is already stated before.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+  non-M-mode resources accessible per the Debug specification cite:[ExtDbg].
+. Triggers must match and fire on a hart only when a supervisor domain with this
+  configuration is active on a hart (and not in M-mode).
```
ved-rivos, line 67: This can also be removed since we already stated that Debug Mode cannot be entered. Firing or/not of triggers depends on context switching of the triggers by the RDSM.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+. The RDSM must be able to context switch all triggers even when `d-mode` = 1.
+  This functional change allows the RDSM to remain in control of external debug
+  for supervisor domains when it is not under external debug itself.
+
+The following figure is non-normative and is intended to illustrate the use of
+the `msdcfg.sdedbgalw` control - the normative specification will be in the
+specification for external debug security cite:[ExtDbgSec].
```
ved-rivos, line 74: This picture can also be removed.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+`msdcfg.sdetralw`.
+
+When M-mode external trace is disabled, whether execution at privilege modes
+less than `M-mode` may be traced by an external trace tool depends on the
+configuration held in `msdcfg.sdetrcalw`, as described below:
+
+When `msdcfg.sdetrcalw` = 0, external trace is disallowed for the SD. All trace
+output from the trace encoder module are disabled while the SD is active on a
+per-hart basis.
+
```
ved-rivos, line 93: "while SD is.." can be removed. Can just state "When `msdcfg.sdetrcalw` = 0, external trace of the SD is disallowed".

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+To enforce the above controls specified by this extension, the following
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
+
+. The enable control for external trace driven by the external trace module is
+  expected to be established by the platform root-of-trust following RISC-V
+  Security Model recommendations SR_GEN_007 and SR_GEN_012. When the control is
+  in cleared state, the hart should not be able to generate trace.
```
ved-rivos, line 105: Same comment as previous section.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+. Per the Efficient trace specification cite:[ETrc], the side-band `halted`
+  signal being asserted, stops subsequent tracing from the hart. On this signal
+  being deasserted, the encoder can start tracing again. Implementation of this
+  `halted` side-band signal is required to support external tracing with
+  supervisor domains. The state of the `halted` side-band signal shall be gated
+  by the external tracing enable control for M-mode.
```
ved-rivos, line 111: This detail is not required. Suffices to say trace is disallowed for this section.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+.. entry to Debug mode
+.. entry to `M-mode` privilege
+.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
+. When extracing tracing is disabled, tracing should be started (via deasserting
+  `halted`) on:
+.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0
+
+When external tracing is enabled (and authorized), there are no restrictions and
+all privilege modes (inclusive of `M-mode` and any supervisor domains managed by
+the RDSM) are traceable. In this case, tracing shall be enabled (via asserting
```
ved-rivos, line 123: all privilege modes (inclusive of `M-mode`  -> this is not right. This section only talks about SD and should not make a statement that enabling tracing for SD enables tracing of M-mode.

chapter3.adoc
```diff
@@ -134,12 +134,21 @@ after writing `mttp`.
...
+  `msdcfg.sdedbgalw` bit.
+. External-trace-allowed state for a supervisor domain is managed via the
+  `msdcfg.sdetrcalw` bit.
```
rsahita, line 146:
```suggestion
. `Smsdedbg` specifies the `msdcfg.sdedbgalw` bit to manage external-debug for a supervisor domain.
. `Smsdetrc` specifies the `msdcfg.sdetrcalw` bit to manage external-trace for a supervisor domain.
```

chapter3.adoc
rsahita: ack

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+less than `M-mode` may be debugged by an external debugger depends on the
+configuration held in `msdcfg.sdedbgalw`, as described below:
+
+When `msdcfg.sdedbgalw` = 0, external debug is disallowed for the supervisor
+domain. Abstract commands and halt request from the debug module are suppressed
+and stay pending while the supervisor domain is active, on a per-hart basis.
+Other debug operations (current or defined in the future) must similarly be kept
+pending while the supervisor domain is active on a hart. Triggers for the
+supervisor domain must be controlled by the RDSM to prevent any leakage of
+information across security domains.
```
rsahita, line 49:
```suggestion
When `msdcfg.sdedbgalw` is 0:

* Access by external debuggers to the memory and/or state of the supervisor domain is disallowed.
* Entry to Debug Mode from a supervisor domain is disallowed.

```

chapter8.adoc
rsahita: sg

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+
+To enforce the above controls specified by this extension, the following
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
+
+. The enable control for external debug driven by the external debug module is
+  expected to be established by the platform root-of-trust following RISC-V
+  Security Model recommendations SR_GEN_007 and SR_GEN_012. When the control is
+  in cleared state, the hart should not be able to enter external debug.
```
rsahita, line 62:
```suggestion
. External debug of M-mode can be enabled only by the HW RoT of the RDSM
 (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012)..
```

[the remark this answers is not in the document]
rsahita: ack

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+be traced by an external trace tool irrespective of the configuration held in
+`msdcfg.sdetralw`.
+
+When M-mode external trace is disabled, whether execution at privilege modes
+less than `M-mode` may be traced by an external trace tool depends on the
+configuration held in `msdcfg.sdetrcalw`, as described below:
+
+When `msdcfg.sdetrcalw` = 0, external trace is disallowed for the SD. All trace
+output from the trace encoder module are disabled while the SD is active on a
+per-hart basis.
```
rsahita, line 92:
```suggestion
When `msdcfg.sdetrcalw` = 0, external trace of the SD is disallowed.
```

chapter8.adoc
rsahita: ack

chapter8.adoc
rsahita: ack

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+  configuration is active on a hart (and not in M-mode).
+. The RDSM must be able to context switch all triggers even when `d-mode` = 1.
+  This functional change allows the RDSM to remain in control of external debug
+  for supervisor domains when it is not under external debug itself.
+
+The following figure is non-normative and is intended to illustrate the use of
+the `msdcfg.sdedbgalw` control - the normative specification will be in the
+specification for external debug security cite:[ExtDbgSec].
 
 [caption="Figure {counter:image}: ", reftext="Figure {image}"]
```
rsahita, line 76:
```suggestion
```

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
-[title= "External Debug for Supervisor Domain", id=Smsdedbg_img]
+[title= "External Debug opt-in for Supervisor Domain", id=Smsdedbg_img]
 image::images/Smsdedbg.png[]
```
rsahita, line 79:
```suggestion
```

chapter8.adoc
rsahita, line 105:
```suggestion
. External trace of M-mode can be enabled only by the HW RoT of the RDSM
 (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
```

rsahita: ack

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+  by the external tracing enable control for M-mode.
+. When external tracing is disabled, tracing should be stopped (via asserting
+  `halted`) on:
+.. entry to Debug mode
+.. entry to `M-mode` privilege
+.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
+. When extracing tracing is disabled, tracing should be started (via deasserting
+  `halted`) on:
+.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0
+
```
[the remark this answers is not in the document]
rsahita: I think it will be good to put these requirements (i can remove the reference to `halted`) as I address the above comment.

chapter8.adoc
rsahita: sorry I meant when M-mode external tracing is enabled - anyway - this comment is not needed any more I think.

chapter8.adoc
```diff
@@ -1,37 +1,124 @@
...
+When external tracing is enabled (and authorized), there are no restrictions and
+all privilege modes (inclusive of `M-mode` and any supervisor domains managed by
+the RDSM) are traceable. In this case, tracing shall be enabled (via asserting
+`halted`) on entry to Debug mode, and stopped on resumption from Debug Mode.
```
rsahita, line 124:
```suggestion
```

chapter3.adoc
```diff
@@ -134,12 +134,19 @@ after writing `mttp`.
...
-`msdcfg.SDEDBALW` bit. Details of `Smsdia` and `Smsdedbg` are described in
-their respective sections in this specification.
+capabilities of the supervisor domain when associated with a hart. +
+The following extensions use the `msdcfg` register to specify additional
+configuration for supervisor domains:
+
+. `Smsdia` uses `msdcfg.SDICN` to specify the active configuration for
+  the supervisor domain interrupt controller associated with the hart.
+. `Smsdedbg` specifies the `msdcfg.sdedbgalw` bit to manage external-debug for a supervisor domain.
+. `Smsdetrc` specifies the `msdcfg.sdetrcalw` bit to manage external-trace for a supervisor domain.
```
rsahita, line 144:
```suggestion
. `Smsdedbg` specifies the `msdcfg.sdedbgalw` bit to manage 
  external-debug for a supervisor domain.
. `Smsdetrc` specifies the `msdcfg.sdetrcalw` bit to manage 
  external-trace for a supervisor domain.
```

chapter8.adoc
```diff
@@ -1,37 +1,106 @@
...
+isolation:
+
+. External trace of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
+. Per the Efficient trace specification cite:[ETrc], the side-band `halted`
+  signal being asserted, stops subsequent tracing from the hart. On this signal
+  being deasserted, the encoder can start tracing again. Implementation of this
+  `halted` side-band signal is required to support external tracing with
+  supervisor domains. The state of the `halted` side-band signal shall be gated
+  by the external tracing enable control for M-mode.
```
rsahita, line 97:
```suggestion
```

chapter8.adoc
rsahita: ack

chapter8.adoc
```diff
@@ -1,37 +1,98 @@
...
+. External trace of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
+. When external tracing is disabled, tracing should be stopped (via asserting
+  `halted`) on:
+.. entry to Debug mode
+.. entry to `M-mode` privilege
+.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
+. When extracing tracing is disabled, tracing should be started (via deasserting
+  `halted`) on:
+.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0
```
rsahita, line 97:
```suggestion
. When external tracing is disabled, tracing must be stopped on:
.. entry to Debug mode
.. entry to `M-mode` privilege
.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
. When external tracing is disabled, tracing must be started on:
.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 1
```

chapter8.adoc
```diff
@@ -1,37 +1,98 @@
...
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
+
+. External debug of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
+. External debug must be able to transition the hart to Debug Mode and access
+  supervisor domain memory and state. In this context, "state" includes all
+  non-M-mode resources accessible per the Debug specification cite:[ExtDbg].
+. Triggers must match and fire on a hart only when a supervisor domain with this
+  configuration is active on a hart (and not in M-mode).
```
rsahita, line 63:
```suggestion
. External debug must be able to access supervisor domain memory and/or state. 
  In this context, "state" includes all supervisor domain resources accessible per the 
  Debug specification cite:[ExtDbg].
. Entry to Debug Mode from a supervisor domain is allowed.
```

chapter8.adoc
rsahita: ack

chapter8.adoc
rsahita: fixed - i kept the above two points in the requirements section below:
- Access by external debuggers to the memory and/or state of the supervisor domain is allowed.
-Entry to Debug Mode from a supervisor domain is allowed.

chapter8.adoc
rsahita: added the 2 points in the requirements below.

chapter8.adoc
```diff
@@ -1,37 +1,95 @@
...
+configuration held in `msdcfg.sdedbgalw`, as described below:
+
+When `msdcfg.sdedbgalw` is 0:
+
+* Access by external debuggers to the memory and/or state of the supervisor domain is disallowed.
+* Entry to Debug Mode from a supervisor domain is disallowed.
+
+
+When `msdcfg.sdedbgalw` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for such a supervisor domain on a per-hart basis. +
```
rsahita, line 50:
```suggestion
`M-mode` is allowed for such a supervisor domain on a per-hart basis, 
with the additional requirements listed below. +
```

chapter8.adoc
```diff
@@ -1,37 +1,95 @@
...
+`msdcfg.sdetralw`.
+
+When M-mode external trace is disabled, whether execution at privilege modes
+less than `M-mode` may be traced by an external trace tool depends on the
+configuration held in `msdcfg.sdetrcalw`, as described below:
+
+When `msdcfg.sdetrcalw` = 0, external trace of the supervisor domain is disallowed.
+
+When `msdcfg.sdetrcalw` = 1 then external trace of privilege modes less than
+`M-mode` shall be allowed for the SD on a per hart basis.
```
rsahita, line 80:
```suggestion
`M-mode` shall be allowed for the SD on a per hart basis, with the
additional requirements listed below.
```

chapter8.adoc
```diff
@@ -1,37 +1,97 @@
...
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
+
+. External trace of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
+. When external tracing is disabled, tracing must be stopped on:
+.. entry to Debug mode
+.. entry to `M-mode` privilege
+.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
+. When external tracing is disabled, tracing must be started on:
```
AoteJin, line 95: Is `disabled` a typo? 
However, even if it intends to mean `enabled` here, the tracing does not necessarily start because it is additional controlled by `trTeEnable`

rsahita: actually, that second point is not needed any more. removing it.

chapter8.adoc
```diff
@@ -1,37 +1,97 @@
...
+.. Entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 1
```
rsahita, line 96:
```suggestion
```

AoteJin (approved):
The wording might be improved in the long run, but the general idea looks good to me. Thanks!

chapter8.adoc
```diff
@@ -1,37 +1,95 @@
...
+
+
+When `msdcfg.sdedbgalw` = 1 then external debug of privilege modes less than
+`M-mode` is allowed for such a supervisor domain on a per-hart basis, 
+with the additional requirements listed below. +
+
+To enforce the above controls specified by this extension, the following
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
```
ved-rivos, line 56: To enforce the above controls specified by this extension, the following
functional and security requirements must be met by the external secure debug
system cite:[ExtDbgSec]:

chapter8.adoc
```diff
@@ -1,37 +1,95 @@
...
+When `msdcfg.sdetrcalw` = 0, external trace of the supervisor domain is disallowed.
+
+When `msdcfg.sdetrcalw` = 1 then external trace of privilege modes less than
+`M-mode` shall be allowed for the SD on a per hart basis, with the
+additional requirements listed below.
+
+To enforce the above controls specified by this extension, the following
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
```
ved-rivos, line 87: To enforce the above controls specified by this extension, the following
functional and security requirements must be met by the external secure debug
system cite:[ExtDbgSec]:

chapter8.adoc
```diff
@@ -1,37 +1,95 @@
...
+
+. External trace of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
+. When external tracing is disabled, tracing must be stopped on:
+.. entry to Debug mode
```
ved-rivos, line 92: Suggesting placing second bullet under the clause "When `msdcfg.sdetrcalw` = 1". Otherwise the context for "when external tracing is disabled" is lost.

chapter8.adoc
rsahita, line 87:
```suggestion
To enforce the above controls specified by this extension, the following
functional and security requirements must be met by the external debug security
system cite:[ExtDbgSec]:
```

chapter8.adoc
```diff
@@ -1,37 +1,95 @@
...
+functional and security requirements must be met by the external secure debug
+system cite:[ExtDbgSec] to meet the security objectives of supervisor domain
+isolation:
+
+. External debug of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
+. External debug must be able to access supervisor domain memory and/or state. 
+  In this context, "state" includes all supervisor domain resources accessible per the 
+  Debug specification cite:[ExtDbg].
+. Entry to Debug Mode from a supervisor domain is allowed.
```
rsahita, line 63:
```suggestion
```

chapter8.adoc
```diff
@@ -1,37 +1,95 @@
...
+When M-mode external trace is disabled, whether execution at privilege modes
+less than `M-mode` may be traced by an external trace tool depends on the
+configuration held in `msdcfg.sdetrcalw`, as described below:
+
+When `msdcfg.sdetrcalw` = 0, external trace of the supervisor domain is disallowed.
+
+When `msdcfg.sdetrcalw` = 1 then external trace of privilege modes less than
+`M-mode` shall be allowed for the SD on a per hart basis, with the
+additional requirements listed below.
+
```
rsahita, line 83:
```suggestion

. When external tracing is disabled, tracing must be stopped on:
.. entry to Debug mode
.. entry to `M-mode` privilege
.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.

```

chapter8.adoc
rsahita, line 56:
```suggestion
To enforce the above controls specified by this extension, the following
functional and security requirements must be met by the external secure debug
system cite:[ExtDbgSec]:
```

chapter8.adoc
```diff
@@ -1,37 +1,99 @@
...
+functional and security requirements must be met by the external debug security
+system cite:[ExtDbgSec]:
+
+. External trace of M-mode can be enabled only by the HW RoT of the RDSM
+ (See RISC-V Security model requirements SR_GEN_007 and SR_GEN_012).
+. When external tracing is disabled, tracing must be stopped on:
+.. entry to Debug mode
+.. entry to `M-mode` privilege
+.. entry to privilege less than `M-mode` with `msdcfg.sdetrcalw` = 0.
+
```
rsahita, line 99:
```suggestion
```

PR #39, opened 2024-03-05
Add CBQRI support for supervisor domains

ved-rivos:


chapter9.adoc
```diff
@@ -0,0 +1,408 @@
...
+[wavedrom, , ]
+....
+{reg: [
+  {bits:  6, name: 'SDICN'},
+  {bits:  1, name: 'SSM'},
+  {bits:  1, name: 'SDEDBGALW'},
+  {bits:  4, name: 'SRL'},
+  {bits:  4, name: 'SML'},
+  {bits:  4, name: 'SQRID'},
+  {bits:  4, name: 'MQRID'},
```
eckhard-delfs-qualcomm, line 159: Question: is the reason to define dedicated fields for SQRID and MQRID that MQRID shall be automatically conveyed as soon as the hart switches into M-mode? In other word, MQRID is supposed to be tied to a fixed value (different from any SQRID), whereas SQRID is intended to be reprogrammed upon switching to a different SD? If so, a clarification would help.

The mttp register introduces SDIDs to differentiate SDs.  There is no definition for an _MDID_ while M-mode is active. It does not seem to be consistent to have different definitions (SQRID+MQRID vs SDID only).

ved-rivos: > Question: is the reason to define dedicated fields for SQRID and MQRID that MQRID shall be automatically conveyed as soon as the hart switches into M-mode? 

The QRID, RCID, and MCID are carried with each request made by the hart to the corresponding shared resource. The shared resource may be a cache, a memory, a TLB, execution resources within a core, power rails, etc. When execution is with M-mode privilege, the MQRID is conveyed with such requests.

> If so, a clarification would help.

Will add.

> The mttp register introduces SDIDs to differentiate SDs. There is no definition for an MDID while M-mode is active. It does not seem to be consistent to have different definitions (SQRID+MQRID vs SDID only).

That is intentional. The mttp register holds the root of the memory tracking table and the SDID is used to tag translations to more efficiently synchronize updates to the MTT with the hart execution. There is no MTT active for M-mode execution.

chapter9.adoc
```diff
@@ -0,0 +1,408 @@
...
+access counters associated with the corresponding `QRID`.
+
+By default all resources in the capacity and bandwidth controllers may be
+allocated using any of the QRI. The controllers may optionally support
+reservation of resources for use by a QRI. When such reservation is supported
+the capacity or bandwidth reserved for a QRI may only be used by requests that
+have the corresponding `QRID`. Supporting resource reservation capability allows
+effective partitioning of the shared resources among SDs sharing the resources.
+
+To optionally support capacity reservation to a QRI, the capacity controllers
```
eckhard-delfs-qualcomm, line 293: Suggestion: To reserve capacity to a QRI, the capacity controllers optionally support a new operation...

chapter9.adoc
eckhard-delfs-qualcomm: > The QRID, RCID, and MCID are carried with each request made by the hart to the corresponding shared resource. The shared resource may be a cache, a memory, a TLB, execution resources within a core, power rails, etc. When execution is with M-mode privilege, the MQRID is conveyed with such requests.
> 
> > If so, a clarification would help.
> 
> Will add.

Thanks!

> > The mttp register introduces SDIDs to differentiate SDs. There is no definition for an MDID while M-mode is active. It does not seem to be consistent to have different definitions (SQRID+MQRID vs SDID only).
> 
> That is intentional. The mttp register holds the root of the memory tracking table and the SDID is used to tag translations to more efficiently synchronize updates to the MTT with the hart execution. There is no MTT active for M-mode execution.

SDs can be run in bare mode without active MTT....

ved-rivos: > SDs can be run in bare mode without active MTT....

When executing in a SD, MTT is always active and there is no way to disable it. However, the memory protection provided by MTT may be effectively disabled by setting the Mode to Bare - a Bare MTT is still an MTT. To draw an analogue, when V=1 (execution in a virtual machine), two-stage address translation is always in effect though either of the stages can be effectively disabled by setting the mode to Bare.

ved-rivos: Included this note:

> The `MQRID` is typically selected once at startup by the RDSM to designate the
> QRI utilized by the RDSM itself. Conversely, the `SQRID` may be updated during
> the process of switching SDs, to denote the QRI allocated to that specific SD by
> the RDSM.

chapter9.adoc
ved-rivos: Updated.

chapter9.adoc
```diff
@@ -0,0 +1,415 @@
...
+     specified in <<SDQRI>> to provide this capability.
+
+* *Associating a QRI ID (QRID) with requests originating from SDs and RDSM.*
+
+  ** The `QRID`, along with the `RCID` and `AT`, is used to identify the
+     resource allocation configurations in a capacity or bandwidth controller.
+     The `QRID`, along with the `MCID`, is used to identify the ID of the
+     counter used to monitor resource usage in a capacity or bandwidth
+     controller. An ISA extension named Smqosid is specified in <<SMQOSID>>
+     to provide this capability.
```
rsahita, line 48: @ved-rivos  this text is repeated below at lines 63-67

chapter9.adoc
```diff
@@ -0,0 +1,415 @@
...
+range of values for `RCID` and `MCID`.
+
+.`msdcfg` register
+
+[wavedrom, , ]
+....
+{reg: [
+  {bits:  6, name: 'SDICN'},
+  {bits:  1, name: 'SSM'},
+  {bits:  1, name: 'SDEDBGALW'},
```
rsahita, line 155: we should shift fields down by 1 bit to use bit 8 for `SDETRCALW` (ext trace allow) in `msdcfg`

rsahita (commented):
couple comments

chapter9.adoc
eckhard-delfs-qualcomm: @ved-rivos : Thanks for adding the note. Looks good to me.

Regarding above ID discussion: the point I was trying to make: Smqosid introduces dedicated QRIDs for M-mode and non-M-mode in order to identify resource allocation configurations and monitor resource usage for _any_ type of CPU transaction. The introduction of MQRID is justified by the fact that for suitable CBQRI support shared resource/capacity controllers need to differentiate M-Mode transactions from those originating from any supervisor domain. Given that MQRID is typically a value that is fixed at startup (see above), the MQRID does not convey more information than the transaction privilege level itself. In other words: conveying MQIRD is equivalent to conveying M-mode privilege level.
On the other hand, SmMTT does not define an M-mode ID (as opposed to SDID) to tag M-mode transactions. The latter does not imply any need for applying MTT checks in M-mode, it only indicates that SmMTT does not consider M-mode transactions for the scope of supervisor domains. But with CBQRI, a distinction SQRID vs MQRID is introduced. Maybe there is some subtlety that I have missed so far.

ved-rivos: While the QRID, RCID, and MCID are intended to be carried with the request, the SDID is not an ID that is intended to be carried with requests. It is also a hart local ID used to identify the SD executing on a hart and not required to be globally unique across all SDs.

rsahita (approved):

rsahita (approved):

rsahita:
@eckhard-delfs-qualcomm are you ok with the updates?

PR #41, opened 2024-03-22
Update to fig1 per issue 18

rsahita:
showing a supervisor domain with only U mode.


rsahita:
cc @gagachang

gagachang:
LGTM, Thanks @rsahita !

PR #52, opened 2024-06-11
Adjust written value for SDIDMAX to match mttp register diagrams

grg-haas:
Registers 1 and 2, for 32-bit and 64-bit `mttp` respectively, show a 6-bit field for `SDID (WARL)`. However, the text describing `SDIDMAX` specifies that the 32-bit value is 8 and the 64-bit value is 16, which both clearly cannot fit into a 6-bit field. Clarify the text to note that `SDIDMAX` should be 6 for both the 32-bit and 64-bit `mttp` registers.


rsahita (approved):

rsahita:
thanks!

PR #53, opened 2024-06-19
Correct signs and explicitly clarify MTTL2 table sizes

grg-haas:
This change mainly clarifies the `MTTL2` table sizes for various SMMTT modes and PAWs. As stated, it was unclear whether the specified table sizes apply when the PAW changes  (i.e. `Smmtt46[rw]` has a table size of `2^(PAW-23)`, `Smmtt34[rw]` has a table size of `2^(PAW-22)`) or when the `RW` mode changes (i.e. `Smmtt46` and `Smmtt34` have a table size of `2^(PAW-23)`, `Smmtt46rw` and `Smmtt34rw` have a table size of `2^(PAW-22)`). Based on the diagrams later in the spec, I believe the latter interpretation is correct so lets make it explicit.

Also, the inequality signs for PAW specifications were incorrect so this commit fixes this as well

### DIFF
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -109,12 +109,12 @@ Implementations are not required to support all defined `MODE` settings when
 Instead, the fields of `mttp` are `WARL` in the normal way, when so indicated.
 
 The `MTTPPN` refers to an `MTTL3` table or an `MTTL2` table based on physical
-address width (`PAW`). For 56 \<= `PAW` < 46, `MTTL3` table must be of size
+address width (`PAW`). For 56 >= `PAW` > 46, `MTTL3` table must be of size
 `2^(PAW-43)` bytes and naturally aligned to that sized byte boundary. For 46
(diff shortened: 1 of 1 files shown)


grg-haas:
This actually brings up another issue, one which I think may be a bug in the spec. Consider that `Smmtt34rw` would have an MTTL2 of size `2^(34-22) = 4096` bytes, or exactly the size of a page. However, `Smmtt34`'s `MTTL2` would then have a size of `2^(34-23) = 2048` bytes, which is only half a page. It would then not actually always be possible to store the **PPN** of this `MTTL2` table in the `mttp` register, since there are two half-pages per PPN. Should there be an extra line in the spec which indicates that `MTTL2` should be at least page aligned?

grg-haas:
I guess this also makes the following line fairly confusing: "In these modes, the lowest two bits of the physical page number (`MTTPPN`) in `mttp` always read as zeros". This would imply alignment of at least 16384 bytes? Also its unclear what specific modes "these modes" is referring to.

rsahita (approved):

rsahita:
thanks!

Issue #49, opened 2024-05-27
MTT implicit accesses should be treated as M-mode accesses for purposes of PMP checking


kasanovic:
Also, should add normative discussion of similar issues as occur for implicit accesses in virtual memory translation (PMAs, endianness, etc).

rsahita:
I added text in PR #48 - will work on adding any other considerations

rsahita:
@kasanovic ptal at PR #58

rsahita:
closing - addressed in PR #58

PR #58, opened 2024-07-08
described MTT lookup process

rsahita:
addressing missing MTT lookup process (Issue #49 )


chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
 ], config:{lanes: 1, hspace:1024}}
 ....
 
+=== MTT access permissions lookup process
+
+MTT access permissions for a physical address PA in the context of a
+supervisor domain is ascertained as follows:
+
+1. Let a be `mttp.ppn`x`PAGESIZE`, and let i = LEVELS, where for mode
```
ved-rivos, line 228: replace: a -> _a_

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+Smmtt[34][*], LEVELS = 2 and for Smmtt[46, 56][*rw*] LEVELS = 3; PAGESIZE is
```
ved-rivos, line 229: Smmtt[34][*] -> Smmtt34*
Smmtt[46, 56][*rw*] -> Smmtt[46|56]*

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+2^12; MTT NON_LEAF_PTE_SIZE = 8 bytes. The `mttp` register must be active,
+i.e., the effective privilege mode must be HS-mode, S-mode or U-mode.
```
ved-rivos, line 231: Also include "HS-mode, S-mode or U-mode." -> "S/HS-mode, U-mode, VS-mode, or VU-mode." or could say "not M".

chapter4.adoc
eckhard-delfs-qualcomm: I thought LEVELS=3 only applies to Smmtt56.

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+
+2. Let mpte be the value of the MTT PTE at address a + pa.ppn[i] x
+NON_LEAF_PTE_SIZE. If accessing pte violates a PMA or PMP check, raise
+an access-fault exception corresponding to the original access type.
```
ved-rivos, line 235: The section describing the table layout uses "PA[..]" terminology for the index bits. Use of "pn[0], pn[1]..." in those tables is better. Here "pa.ppn" -> "pa.pn".

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+
+3. If any bits or encodings that are reserved for future standard use are
+set within mpte, stop and raise a page-fault exception corresponding to
```
ved-rivos, line 238: mpte -> _mpte_

ved-rivos: "page-fault" -> "access fault"

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+the original access type.
+
+4. Otherwise, the mpte is valid. If (i=0) or (i=1 and mpte.type is not
+`MTT_L1_DIR`), go to step 5. Otherwise, the mpte is a pointer to the next
+level of the MTT. Let i = i - 1. If i < 0, stop and raise a page-fault
+exception corresponding to the original access type. Otherwise, let
```
ved-rivos, line 244: "page-fault" -> "access fault"

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
 ....
 
+=== MTT access permissions lookup process
+
+MTT access permissions for a physical address PA in the context of a
+supervisor domain is ascertained as follows:
+
+1. Let a be `mttp.ppn`x`PAGESIZE`, and let i = LEVELS, where for mode
+Smmtt[34][*], LEVELS = 2 and for Smmtt[46, 56][*rw*] LEVELS = 3; PAGESIZE is
+2^12; MTT NON_LEAF_PTE_SIZE = 8 bytes. The `mttp` register must be active,
```
eckhard-delfs-qualcomm, line 230: Any reason why we call it NON_LEAF_PTE_SIZE instead of MPTE_SIZE?

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+4. Otherwise, the mpte is valid. If (i=0) or (i=1 and mpte.type is not
+`MTT_L1_DIR`), go to step 5. Otherwise, the mpte is a pointer to the next
+level of the MTT. Let i = i - 1. If i < 0, stop and raise a page-fault
+exception corresponding to the original access type. Otherwise, let
+a = mpte.ppn x PAGESIZE and go to step 2. Note that when mpte.type =
+MTT_L1_DIR, the mpte.ppn field is the value of the mpte.info field.
+
+5. A leaf mpte has been found. Determine if the requested physical memory
+access is allowed per the access permission encoding, AND'ing the MTT
+permissions with the cumulative read, write, execute permissions derived from
```
eckhard-delfs-qualcomm, line 250: What is the purpose of the statement "AND'ing the MTT permissions etc.."? MTT applies to physical accesses only, i.e. at the stage when the 3rd-stage lookup is performed, the lower stages (1/2) already decided to proceed to/block physical access.

ved-rivos, line 250: Why allow for write-only permissions? Why do the ANDing? If the read permission is not available then cause an access fault. If the access is an implicit or explicit write then cause an access fault.

chapter4.adoc
ved-rivos: for 46 and 56 the level should be 3.

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+1st-stage and G-stage translation (if active). If access is not permitted,
```
ved-rivos, line 251: neither stage translation need to be active for MTT checks.

ved-rivos: The steps for processing L1 table should be described in a separate step. It may be useful  to separate the procedure for the three *rw format from the other three format. The tests for reserved encodings in L1 are missing.

ved-rivos: Checks for reserved bits in INFO field bits 63:32 (table 4 seems to have a typo) is missing.

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+stop and raise a page-fault exception corresponding to the original access
+type. If i=1, the mpte.type field (for 1GB pages) and the mpte.info field
+(for 2MB pages) holds the access permissions for superpages(also shown in
+Tables 3 and 4). If i=0, the mpte contains 2 bit access permission encodings
+(or 4 bit for mode `Smmtt[*]rw`) for 4KB pages. The 2/4-bit access permission
+encoding for the PA is held in the ppn referenced by mpte.info and
+indexed via pa.ppn[i]. For modes `Smmtt[*]`, access allowed implies read,
```
ved-rivos, line 258: Since these are bit indices they may want to be called out separately.

chapter4.adoc
```diff
@@ -221,9 +220,54 @@ follows:
...
+3. If any bits or encodings that are reserved for future standard use are
+set within mpte, stop and raise a page-fault exception corresponding to
+the original access type.
+
+4. Otherwise, the mpte is valid. If (i=0) or (i=1 and mpte.type is not
+`MTT_L1_DIR`), go to step 5. Otherwise, the mpte is a pointer to the next
+level of the MTT. Let i = i - 1. If i < 0, stop and raise a page-fault
+exception corresponding to the original access type. Otherwise, let
+a = mpte.ppn x PAGESIZE and go to step 2. Note that when mpte.type =
+MTT_L1_DIR, the mpte.ppn field is the value of the mpte.info field.
```
ved-rivos, line 246: These should go into two different step - one for L1_DIR case and one otherwise.

chapter4.adoc
rsahita, line 228:
```suggestion
1. Let _a_ be `mttp.ppn`x`PAGESIZE`, and let _i_ = LEVELS, where for mode
```

chapter4.adoc
rsahita, line 229:
```suggestion
Smmtt34, LEVELS = 2 and for Smmtt[46 | 56][rw] LEVELS = 3; PAGESIZE is
```

chapter4.adoc
rsahita: at the leaf level, the entry size for MTT entries are 2 bits and 4 bits for Smmtt[*] and Smmtt[*]rw resp.

chapter4.adoc
rsahita, line 231:
```suggestion
i.e., the effective privilege mode must be not-M-mode.
```

chapter4.adoc
rsahita, line 238:
```suggestion
set within _mpte_, stop and raise an access-fault exception corresponding to
```

chapter4.adoc
rsahita, line 244:
```suggestion
4. Otherwise, the _mpte_ is valid. If (_i_=0) or (_i_=1 and _mpte.type_ is not
`MTT_L1_DIR`), go to step 5. Otherwise, the _mpte_ is a pointer to the next
level of the `MTT`. Let _i_ = _i_ - 1. If _i_ < 0, stop and raise an access-fault
exception corresponding to the original access type. Otherwise, let
```

chapter4.adoc
rsahita: for implicit reads, writes that makes sense - for the final PA - the 1st stage and/or the G-stage permissions may disallow writes, but the MTT may allow rw - and the effective permission if any caching is performed should be the AND of the permissions accumulated from 1st, G and MTT.

rsahita: perhaps i need to split implicit vs explicit and specify leaf pte - for implicit reads, writes PA permissions from MTT alone is sufficient to allow/deny - for the final PA - the 1st stage and/or the G-stage permissions may disallow writes, but the MTT may be in a Smmtt[x]rw mode and allow rw for that PA - in that case the effective permission if any caching is performed should be the AND of the permissions accumulated from 1st, G and MTT.

chapter4.adoc
rsahita: i can split the steps out for the two top level modes. agree that neither stage translation needs to be active - hence the "if active" note in parentheses.  

will add note for reserved bits in INFO field (thanks)

chapter4.adoc
rsahita: I didn't parse your comment. there are only 2 cases, leaf or non-leaf - the case for non-L1_DIR is the leaf case which is handled in step 5.

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
+3. If any bits or encodings that are reserved for future standard use are
+set within _mpte_, stop and raise an access-fault exception corresponding to
+the original access type.
+
+4. Otherwise, the _mpte_ is valid. If (_i_=0) or (_i_=1 and _mpte.type_ is not
+`MTT_L1_DIR`), go to step 5. Otherwise, the _mpte_ is a pointer to the next
+level of the `MTT`. Let _i_ = _i_-1. If _i_ < 0, stop and raise an access-fault
+exception corresponding to the original access type. Otherwise, let
+_a_ = _mpte.ppn_ x _PAGESIZE_ and go to step 2. Note that when _mpte.type_ =
+`MTT_L1_DIR`, the _mpte.ppn_ field is the value of the _mpte.info_ field.
```
ved-rivos, line 250: This note is not needed since if it was MTT_L1_DIR we would have branched to step 5.

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
+* if _i_=1, and the _mpte.type_ field specifies it holds the access-permission
+encoding for 1GB pages per <<Smmtt-rw-l2-encoding>> and <<Smmtt-ad-l2-encoding>>
+; go to step 7, else
+
+* if _i_=1, and the _mpte.type_ field specifies it holds the access-permission
+encoding for 2MB pages; the _mpte.info_ field contains the access-permission
+encoding per <<Smmtt-rw-l2-encoding>> and <<Smmtt-ad-l2-encoding>>; go to
+step 7, else
+
+* If _i_=0, the _mpte_ contains a 2-bit entry for mode `Smmtt[34 | 46 | 56]`
```
ved-rivos, line 269: when i is 0, the mpte holds an array of 2 bit or 4-bit entries.

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
 ....
 
+=== MTT access permissions lookup process
+
+MTT access-permissions for a physical address PA in the context of a
+supervisor domain is ascertained as follows:
+
+1. Let _a_ be `mttp.ppn` x _PAGESIZE_, and let _i_ = LEVELS, where for mode
+`Smmtt34[rw]`, _LEVELS_ = 2 and for `Smmtt[46 | 56][rw]` LEVELS = 3; _PAGESIZE_
+is 2^12; MTT _NON_LEAF_PTE_SIZE_ = 8 bytes. The `mttp` register must be active,
```
eckhard-delfs-qualcomm, line 234: 2^12 -> 2<sup>12</sup>  (seems to be the preferred style in RV docs)

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
+encoding for 2MB pages; the _mpte.info_ field contains the access-permission
+encoding per <<Smmtt-rw-l2-encoding>> and <<Smmtt-ad-l2-encoding>>; go to
+step 7, else
+
+* If _i_=0, the _mpte_ contains a 2-bit entry for mode `Smmtt[34 | 46 | 56]`
+that holds the access-permission encodings (or 4-bit entry for mode
+`Smmtt[34 | 46 | 56]rw`) for 4KB pages. The 2/4-bit access-permission encoding
+for the PA is held in the ppn referenced by _mpte.info_ and indexed via
+_pa.pn[i]_. The encodings are specified in <<Smmtt-rw>> and <<Smmtt-ad>> for the
+`Smmtt[34 | 46 | 56]` and `Smmtt[34 | 46 | 56]rw` modes respectively.
```
ved-rivos, line 274: This could be more clearly specified if the *rw was separate into its own section or similar to LEVELS there should be a parameter for 2 or 4 bit that is setup ahead of time and then used in the algorithm to extract the right 2 or 4 bits - this could be described in the form of a shift and a mask.

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
+
+7. Determine if the requested physical memory access is allowed per the
+access-permissions. If access is not permitted, stop and raise a page-fault
```
ved-rivos, line 277: This determination varies based on *rw or not. This step should elaborate precisely how this determination is made. If all encodings of the 2 or 4 bit are not legal then that also needs to be checked.

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
+AND'ed to accumulate the effective read, write, and execute permissions. Thus,
+MTT access-permissions can only further restrict access, and never grant
+permission denied by 1st-stage or G-stage translations.
+====
+
 === Access Enforcement and Fault Reporting
 
-As shown in <<mtt-lookup>>, MTT lookup composes with, but does not require,
+As shown in <<mtt-lookup>>, and described in the MTT lookup process,
+MTT lookup composes with, but does not require,
```
ved-rivos, line 301: pn's are referenced in the algorithm as indexed by the level `i` buit the diagrams show pn as being a single field (e.g. diagram shows pn[56:46] is used as index into L2) but the algorithm says pn[i] - for i=2 this would mean use bit 2 of pn.

chapter4.adoc
ved-rivos: In table 3 - 1G_Allow_r can be changed to 1G_Allow_rx. The descriptions in table  3 should also clarify what is meant by read/write.

ved-rivos: Likewise the PERM description:
"0001b - the 4K page specifies read access is allowed for the domain" -> "0001b - the 4K page specifies read and execute access is allowed for the domain"

chapter4.adoc
ved-rivos: Could the two table formats have been unified as only 2 bits are needed in either case? 
Could it be:
00 - not allowed
01 - read and execute allowed
10 - read and write allowed
11 - read, write, and execute allowed
Then 01 and 10 encoding can just be reserved for non-rw mode but the table layout would be identical.

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
+
+All implicit accesses to the non-leaf memory tracking table data structures in
+this algorithm are performed using width _NON_LEAF_PTE_SIZE_.
+
+[NOTE]
+====
+Implementations may cache any translation (due to 1st stage and/or G-stage
+translation) and access-permissions for the final (leaf) physical address
+being accessed - in this case, the `MTT`-derived access-permissions are
+AND'ed to accumulate the effective read, write, and execute permissions. Thus,
```
ved-rivos, line 293: Smmtt should not prescribe microarchitecture. ANDing with effective read/write/execute permissions will be wrong - consider single stage page table for simplicity. If the TLB has a r, w, and x bit - and say the single stage PTE grants them all but the MTT only grants read/execute permission. With anding it would lead to TLB holding R=1, W=0, X=1. Now if a store hits that TLB entry should implementation raise a page-fault or should it raise an access fault?

chapter4.adoc
```diff
@@ -221,9 +224,81 @@ follows:
...
+
+MTT access-permissions for a physical address PA in the context of a
+supervisor domain is ascertained as follows:
+
+1. Let _a_ be `mttp.ppn` x _PAGESIZE_, and let _i_ = LEVELS, where for mode
+`Smmtt34[rw]`, _LEVELS_ = 2 and for `Smmtt[46 | 56][rw]` LEVELS = 3; _PAGESIZE_
+is 2^12; MTT _NON_LEAF_PTE_SIZE_ = 8 bytes. The `mttp` register must be active,
+i.e., the effective privilege mode must be not-M-mode.
+
+2. Let _mpte_ be the value of the `MTT` table entry at address _a_ + _pa.pn[i]_
```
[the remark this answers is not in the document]
rsahita: will fix

chapter4.adoc
rsahita: reduced the note to the main part:

[NOTE]
====
MTT access-permissions can only further restrict access, and never grant
read, write or execute permission denied by 1st-stage or G-stage translations.
====

chapter4.adoc
rsahita: will fix

chapter4.adoc
rsahita: will be updated in PR revision - ptal

chapter4.adoc
rsahita: will be updated in PR revision - ptal

chapter4.adoc
rsahita: updated

chapter4.adoc
```diff
@@ -2,17 +2,17 @@
...
-widths e.g. 46 bits, the `MTTL3` table is not applicable. In this mode of the
-Smmtt[34, 46, 56]rw, each page is associated with a read and a write access
-permission (2 bits). <<mtt-map-rw>> shows the Physical address (PA) being
-used to index into the `MTT` structure in memory to lookup access
-permissions for the supervisor domain specified in the `MTTL1` entry.
-Intermediate `MTTL3` and `MTTL2` entries are used to allow this structure to
-be sparsely populated.
+widths e.g. 46 bits, the `MTTL3` table is not applicable. Each
+physical-addressable page is associated with an access-permission encoding.
+<<mtt-map-rw>> shows the Physical address (PA) being used to index into the
```
ved-rivos, line 12:
```suggestion
<<mtt-map-rw>> shows the Supervisor Physical address (SPA) being used to index into the
```

chapter4.adoc
```diff
@@ -2,17 +2,17 @@
...
 The `PPN` rooted structure for the MTT is shown below. The structure
 below shows a 56 bit physical address lookup; for lower physical address
-widths e.g. 46 bits, the `MTTL3` table is not applicable. In this mode of the
-Smmtt[34, 46, 56]rw, each page is associated with a read and a write access
-permission (2 bits). <<mtt-map-rw>> shows the Physical address (PA) being
-used to index into the `MTT` structure in memory to lookup access
-permissions for the supervisor domain specified in the `MTTL1` entry.
-Intermediate `MTTL3` and `MTTL2` entries are used to allow this structure to
-be sparsely populated.
+widths e.g. 46 bits, the `MTTL3` table is not applicable. Each
```
ved-rivos, line 10: Could we add a note that the sizes shown for L3 and L2 tables are their maximum size. However only the parts of the table that correspond to memory present in the platform need to be actually available. Could give an example of where if a machine only has 40 bit physical memory as an example to show that only first N entries of L2 are accessed.

chapter4.adoc
```diff
@@ -2,17 +2,17 @@
...
+physical-addressable page is associated with an access-permission encoding.
+<<mtt-map-rw>> shows the Physical address (PA) being used to index into the
+`MTT` structure in memory to lookup access permissions for the supervisor domain
+specified in the `MTTL1` entry. Intermediate `MTTL3` and `MTTL2` entries are
+used to allow this structure to be sparsely populated.
```
ved-rivos, line 15: Smmtt34 is clearly intended to support RV32. Howeve the L2 entries are 8 bytes wide which is wider than RV32 XLEN. Do we want to separate Smmtt34 to only have 4 byte L2 entries?

chapter4.adoc
```diff
@@ -46,57 +46,65 @@ The `ZERO` field must always be 0.
...
+for the supervisor domain (described in the `MTTL1` entry <<mtt-l1-rw>>).
 
 |`2M_PAGES` a|
 _The 32M range of address space is partitioned into 2M pages where each
 page has read/write access allowed/not._
 The `INFO` field 31:0 holds 2 `PERM` bits per 2M address range to indicate
-`access_disallowed` (00b), `read_allowed` (01b), `read_write_allowed` (11b);
-(10b is `reserved`). Bits 32:43 are reserved and must be zero.
+`rwx_disallowed` (00b), `rx_allowed` (01b), `rw_allowed` (10b), `rwx_allowed`
+(11b). Bits 32:43 are reserved and must be zero.
```
ved-rivos, line 105: Indicate that other INFO field bits must be 0? When access is not allowed  it causes access violation.

chapter4.adoc
```diff
@@ -46,57 +46,65 @@ The `ZERO` field must always be 0.
...
 The `INFO` field provides the PPN of the `MTTL1` page.
-Entries of the `MTTL1` page hold a 4-bit `PERM` (permissions) field to indicate
-the access for the supervisor domain (described in the `MTTL1` entry
-<<mtt-l1-rw>>).
+Entries of the `MTTL1` page hold a 4-bit `PERM` field to indicate the access
+for the supervisor domain (described in the `MTTL1` entry <<mtt-l1-rw>>).
 
 |`2M_PAGES` a|
 _The 32M range of address space is partitioned into 2M pages where each
 page has read/write access allowed/not._
```
ved-rivos, line 102: read/write/execute

chapter4.adoc
ved-rivos: rwx_disallowed -> access disallowed?

ved-rivos: It may not be obvious what "rx_allowed" means. Maybe use same language as used for L1 access permission encodings description?

chapter4.adoc
ved-rivos: Having the SPA being show as having 4 pn parts may want to be a separate diagram and it can be referenced from the walk section. The table layout can then be another diagram (which also includes the PA with pn split). Alternately some way to draw attention that "pn[i]" in walk process actually refers to what is shown in this diagram would help.

chapter4.adoc
```diff
@@ -123,107 +145,71 @@ domain
...
+for 16 2M address regions (see <<Smmtt-rw-l2-encoding>>); go to step 7, else
 
-|`2M_PAGES` a|
-_The 64M range of address space is partitioned into 2M pages where each
-page has access allowed/not._
-The `INFO` field bits 31:0 holds 1 bit per 2M address range to indicate
-access disallowed(0b) or allowed (1b). `INFO` field bits 43:32 are
-reserved (must be zero).
-|===
+* If _i_=0, the _mpte_ contains a 4-bit entry that holds the access-permission
```
ved-rivos, line 189: The mpte actually contains 16x4-bit entries. May want to describe this as a shift and mask or say extract bits i:j where i and j are determined by pa[0].

chapter4.adoc
```diff
@@ -123,107 +145,71 @@ domain
...
+encodings for 4KB pages. The 4-bit access-permission encoding for the _pa_ is
+held in _mpte_ and indexed via _pa.pn[i]_. The encodings are specified in
+<<Smmtt-rw-l1-encoding>> table.
 
-`MTTL1` table is populated if 4KiB page confidential pages are required
-for the supervisor domain.
+7. Determine if the requested physical memory access is allowed per the
+access-permissions. If access is not permitted, stop and raise a page-fault
```
ved-rivos, line 195: page-fault -> access-fault.

chapter4.adoc
```diff
@@ -123,107 +145,71 @@ domain
...
-in the page and bits 16:12 are used to select a 2 bit `PERM` field in those 64
-bits. Thus, there are 2 `PERM` bits for each 4 KiB page. The encoding is as
-follows:
+8. The access is allowed per the `MTT` lookup.
 
-* 00b - the 4K page specifies access is *not allowed* for the domain
-* 01b - the 4K page specifies access is *allowed* for the domain
-* 1xb - *reserved* (access causes access violation).
+All implicit accesses to the non-leaf memory tracking table data structures in
+this algorithm are performed using width NON_LEAF_PTE_SIZE.
```
ved-rivos, line 201: This may be problematic for RV32.

chapter4.adoc
rsahita, line 102:
```suggestion
page has read/write/execute access allowed/not._
```

chapter4.adoc
rsahita, line 195:
```suggestion
access-permissions. If access is not permitted, stop and raise an access-fault
```

chapter4.adoc
```diff
@@ -46,57 +46,65 @@ The `ZERO` field must always be 0.
...
+[[Smmtt-rw-l2-encoding]]
 .`MTTL2` Entry Type for Smmtt
 [width="100%",cols="25%,75%",options="header",]
 |===
 |*`MTTL2` Entry `TYPE`* |*Description, `INFO` and `TYPE` field encoding*
 |`1G_disallow` a|
-_The 1G range of address is not allowed for the domain._
+_read, write or execute is not allowed to this 1G address range for the domain._
+The `INFO` field must be 0.
+When configuring 1G ranges, RDSM ensures that 16 `MTTL2` entries, each
```
eckhard-delfs-qualcomm, line 72: Suggestion: rephrase (and subsequent instances) to "RDSM must ensure", since it is about FW responsibility.

eckhard-delfs-qualcomm, line 72: that 16 `MTTL2` entries, each corresponding to 64M of address space
-> that **32** `MTTL2` entries, each corresponding to **32M** of address space

chapter4.adoc
```diff
@@ -123,107 +145,71 @@ domain
...
-_The 64M range of address space is partitioned into 2M pages where each
-page has access allowed/not._
-The `INFO` field bits 31:0 holds 1 bit per 2M address range to indicate
-access disallowed(0b) or allowed (1b). `INFO` field bits 43:32 are
-reserved (must be zero).
-|===
+* If _i_=0, the _mpte_ contains a 4-bit entry that holds the access-permission
+encodings for 4KB pages. The 4-bit access-permission encoding for the _pa_ is
+held in _mpte_ and indexed via _pa.pn[i]_. The encodings are specified in
+<<Smmtt-rw-l1-encoding>> table.
```
eckhard-delfs-qualcomm, line 192:
```suggestion
<<Smmtt-rw-l1-encoding>>.
```

chapter4.adoc
rsahita, line 72:
```suggestion
When configuring 1G ranges, RDSM must ensure that 32 `MTTL2` entries, each
```

chapter4.adoc
```diff
@@ -46,57 +46,65 @@ The `ZERO` field must always be 0.
...
 .`MTTL2` Entry Type for Smmtt
 [width="100%",cols="25%,75%",options="header",]
 |===
 |*`MTTL2` Entry `TYPE`* |*Description, `INFO` and `TYPE` field encoding*
 |`1G_disallow` a|
-_The 1G range of address is not allowed for the domain._
+_read, write or execute is not allowed to this 1G address range for the domain._
+The `INFO` field must be 0.
+When configuring 1G ranges, RDSM ensures that 16 `MTTL2` entries, each
+corresponding to 64M of address space, have identical `TYPE` field values.
```
rsahita, line 73:
```suggestion
corresponding to 32M of address space, have identical `TYPE` field values.
```

chapter4.adoc
rsahita: will make all MTT L2 and L1 PTEs XLEN dependent

chapter4.adoc
rsahita: will fix by making MTT L2, L1 entries XLEN dependent

chapter4.adoc
rsahita: updated to show SPA separately also

chapter4.adoc
rsahita: fixed

rsahita:
Please look at the last build check artifact to review the PDF version - 
link provided for convenience -> https://github.com/riscv/riscv-smmtt/actions/runs/10465986323/artifacts/1830808709

images/fig3b.png
eckhard-delfs-qualcomm, line 1: Figure 8: MTT walk for RV32 shows an address space per entry of 16MB (vs 32 MB in RV64).
Should this be 32MB instead for consistency reasons? Otherwise, the description in Table 3 (MTTL2 Entry Type for Smmtt) does not fit for RV32.
("When configuring 1G ranges, RDSM must ensure that 32 MTTL2 entries, each corresponding to **32M**...")

If so, should pn[2] for RV32 be reduced to 9 bits [33:25] (cf Figure 5)? This would allow to expand pn[1] to 10 bits [24:15] and get rid of L/H bit.

rsahita: issue is if pn(1) is expanded to 10 bits then for 2MB pages, we need 16 combinations of perms, for which we do not have space in the 4 byte L2 PTE with the type field.

chapter4.adoc
```diff
@@ -41,79 +76,119 @@ The following MTT L2/L3 entries accomodate a 56 bit PAW.
...
 
+[caption="Figure {counter:image}: ", reftext="Figure {image}"]
+[title="`MTTL2` entry (XLEN=32)"]
+[wavedrom, ,svg]
+....
+{reg: [
+  {bits:  22, name: 'INFO'},
+  {bits:  4,  name: 'TYPE'},
+  {bits:  1,  name: 'L/H'},
+  {bits:  6,  name: 'ZERO'},
```
eckhard-delfs-qualcomm, line 87:
```suggestion
  {bits:  5,  name: 'ZERO'},
```

images/fig3b.png
eckhard-delfs-qualcomm: @rsahita: Understood. The entry for "2M_Pages" would then also need some update to consider the RV32 case.
Thinking about it, would it be an alternative to switch from 2M to 4M_Pages instead? This size could serve as a common denominator for both RV32 and RV64 and use unified mpte.info[15:0] usage for the permissions.

rsahita: yeah that's better - let me make that change and update the PR.

rsahita: PTAL at the updates @eckhard-delfs-qualcomm

rsahita:
Updated draft -> https://github.com/riscv/riscv-smmtt/actions/runs/10533722659/artifacts/1849308243

eckhard-delfs-qualcomm (commented):
Thanks, looks good to me, just a few minor comments.

rsahita:
Updated draft link -> https://github.com/riscv/riscv-smmtt/actions/runs/10563011302/artifacts/1855303902

PR #71, opened 2024-09-16
QC suggestions for RC1 (some minor clarifications)

eckhard-delfs-qualcomm:
- Removal of "allow/disallow" only option from Smmtt34 
- adding explicit statement how many SDs are supported by SmMTT
- clarification that permission attribute is "disallowed" not allowed
- ambiguity: "MTT structure accesses" -> MTT checker accesses to MTT structures
- clarification to informative statement on MTT lookups in section 4.4:  Chapter 2 states that when Smmtt is implemented, MTT and e(PMP) are **always** active. Hence the phrases "while active / no longer active" can be misleading.

### DIFF
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -113,8 +113,7 @@ every bit position in the `SDID` field, then reading back the value in `mttp`
 to see which bit positions in the `SDID` field hold a one. The
 least-significant bits of `SDID` are implemented first: that is, if `SDIDLEN` >
 0, `SDID`[`SDIDLEN`-1:0] is writable. The maximal value of `SDIDLEN`, termed
-`SDIDMAX`, is 6 for `Smmtt[34 | 46 | 56]` and consequently allows for up to
-64 supervisor domains.
+`SDIDMAX`, is 6 for `Smmtt[34 | 46 | 56]`.
 
 The `mttp` register is considered active for the purposes of the physical
 address protection algorithm unless the effective privilege mode is `M`.
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -54,7 +54,7 @@ supervisor domains beyond the physical memory protection scheme described in
 Section 3.7 of the RISC-V privileged architecture specification cite:[ISA]. In
 this case, the remaining fields (`SDID`, `PPN`) in `mttp` must be set to
 zeros, else generate a fault. When `XLEN=32`, the other valid setting for
-`MODE` is `Smmtt34` to support allow/disallow and read-write-execute
+`MODE` is `Smmtt34` to support read-write-execute
 access permissions for 34-bit system physical addresses.
 
 When `XLEN=64`, other than `BARE`, the other valid settings for `MODE` are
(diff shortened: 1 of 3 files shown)


chapter3.adoc
```diff
@@ -113,7 +113,8 @@ every bit position in the `SDID` field, then reading back the value in `mttp`
 to see which bit positions in the `SDID` field hold a one. The
 least-significant bits of `SDID` are implemented first: that is, if `SDIDLEN` >
 0, `SDID`[`SDIDLEN`-1:0] is writable. The maximal value of `SDIDLEN`, termed
-`SDIDMAX`, is 6 for `Smmtt[34 | 46 | 56]`.
+`SDIDMAX`, is 6 for `Smmtt[34 | 46 | 56]` and consequently allows for up to
+64 supervisor domains.
```
SiFiveHolland, line 117: My reading of the spec is that SDIDs are intended to work like ASIDs: they can be reused if you run out, at the cost of requiring a full-address-space fence when repurposing an SDID for a different supervisor domain (either at startup or during the context switch). So there is effectively no limit on the number of active supervisor domains. I have opened #76 to clarify this.

rsahita: thats correct - see the comment on the use of SDIDs in that issue #76

rsahita, line 117:
```suggestion
`SDIDMAX`, is 6 for `Smmtt[34 | 46 | 56]`.
```

chapter4.adoc
```diff
@@ -149,7 +149,7 @@ RV64 and <<mtt-l1-rv32>> for RV32.
 |`2M_PAGES` a|
 This encoding applies only to RV64.
 _The 32 MiB range of address space is partitioned into 16 2 MiB pages where each
-page has read/write/execute access allowed/not specified via the `INFO` field._
+page has read/write/execute/access disallowed/not specified via the `INFO` field._
```
rsahita, line 152:
```suggestion
page has read/write/execute access specified via the `INFO` field._
```

chapter4.adoc
```diff
@@ -159,7 +159,7 @@ reserved and must be zero.
 |`4M_PAGES` a|
 This encoding applies only to RV32.
 _The 32 MiB range of address space is partitioned into 8 4 MiB pages where each
-page has read/write/execute access allowed/not specified via the `INFO` field._
+page has read/write/execute/access disallowed/not specified via the `INFO` field._
```
rsahita, line 162:
```suggestion
page has read/write/execute access specified via the `INFO` field._
```

chapter4.adoc
```diff
@@ -347,8 +347,8 @@ batch optimized via the use of the sequence `SFENCE.W.INVAL`, `MINVAL.SPA` and
 
 [NOTE]
 ====
-MTT lookups that began while `mttp` was active are not required to complete or
-terminate when `mttp` is no longer active, unless a `MFENCE.SPA` instruction
+Ongoing MTT lookups are not required to complete or terminate when `mttp` is 
+effectively disabled, unless a `MFENCE.SPA` instruction
```
rsahita, line 351:
```suggestion
MTT lookups that began while `mttp` was active are not required to complete or
terminate when `mttp` is no longer active, unless a `MFENCE.SPA` instruction
```

rsahita: reverting this change - following the same language as the priv spec for vs and g stage lookups.

rsahita (approved):
edits look fine to me with some changed merged in.

PR #79, opened 2024-10-01
clarify behavior of invalid SDICN

### LINKED ISSUE #59: Question about siselect when SDICN is zero
Current Smsdia (section 6.1) states:

```
When SDICN is not the number of an implemented interrupt controller, or is zero then the following
rules apply to all privilege modes:

• All siselect values that access IMSIC registers are reserved.
```

What does "reserved" mean ?
Shouldn't it raise an illegal instruction exception in this case ?

Recommended changes if it should raise an illegal instruction exception:
```
When SDICN is not the number of an implemented interrupt controller, or is zero then the following
rules apply to all privilege modes:

• When siselect is a number in the range 0x70–0xFF (to access IMSIC registers), attempts to access
sireg raises an illegal instruction exception.
```

Therefore, RDSM can emulate an IMSIC when SDICN is set to 0.

ved-rivos:
closes #59

### DIFF
--- a/chapter7.adoc
+++ b/chapter7.adoc
@@ -124,9 +124,10 @@ When `SDICN` is not the number of an implemented interrupt controller, or is
 zero then the following rules apply to all privilege modes:
 
 * The S-level external interrupt pending signal indicated in `mip.SEIP` is 0.
-* All non-custom values of `siselect` designate an _inaccessible_ register. When
-  `siselect` has the number of an _inaccessible_ register, attempts from M-mode
-  or HS-mode to access `sireg` raise an illegal instruction exception.
+* All non-custom values of `siselect` that access IMSIC registers designate an
+  _inaccessible_ register. When `siselect` holds the number of an _inaccessible_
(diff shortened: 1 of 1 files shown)


chapter7.adoc
```diff
@@ -124,13 +124,16 @@ When `SDICN` is not the number of an implemented interrupt controller, or is
 zero then the following rules apply to all privilege modes:
 
 * The S-level external interrupt pending signal indicated in `mip.SEIP` is 0.
-* All `siselect` values that access IMSIC registers are reserved.
+* All non-custom values of `siselect` designate an _inaccessible_ register. When
```
SiFiveHolland, line 127: Is this correct when there are users of Sscsrind other than AIA?

ved-rivos: It should have been a bit more precise. "All non-custom values that access IMSIC registers"

ved-rivos: updated.

rsahita (approved):
lgtm

Issue #76, opened 2024-09-17
Questions about SDIDs


SiFiveHolland:
Due to the fact that SDIDLEN is small (allowed to be as low as 0), and the only standard use of SDIDs is for fences, it appears that SDIDs are intended to be a short-lived identifier, not remaining stable over the life of a supervisor domain.

* Is this correct? If so, I would suggest the specification explicitly state that SDIDs are not a stable identifier.
* Is it expected that the same supervisor domain may have different SDIDs across harts or I/O MTTs?
* Is it allowable for the platform to tag bus transactions (e.g. NS bit or equivalent) with the current SDID or some subset of it?

rsahita:
that is not the case - SDID is a hart-local identifier, in the sense, the RDSM can activate a SDID (and MTT) on a specific hart which another hart has a different SDID (and MTT) active. For most typical usages, the SDID will remain stable over the life of the supervisor domain.  It is allowed for a platform to tag bus transactions with current SDID (and other IDs) - for e.g. to drive downstream crypto context selection (but that is implementation-specific).

I can add this into the non-normative/informative section of the spec.

ved-rivos:
SDID should not attempt to solve the ID on the bus problem. There is a fast track extension underway to define a world ID for that purpose: https://drive.google.com/drive/u/0/folders/1PGDrAZoAc6qin1jqxl6TCDFWIJtP5I56

rsahita:
@SiFiveHolland Any other changes to PR #85 ?

SiFiveHolland:
I am assuming from your response that statement 2 is correct (and statement 1 is not). So we shouldn't suggest mapping the SDID to anything outside the hart.

PR #85, opened 2024-10-01
clarification on SDID as a local identifer

rsahita:
to address issue #76

### DIFF
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -10,9 +10,7 @@ associated with the supervisor domain. For most typical usages, the SDID will
 remain stable over the life of the supervisor domain. However, the SDID can be
 changed at the cost of fencing to flush any prior cached state. The SDID is
 unique only to a hart, hence, SDID must not be used to tag downstream
-transactions that are globally visible - implementations may map the SDID from
-the initiator to other (global) identifiers but that mapping is
-implementation-specific and out of scope of this specification.
+transactions that are globally visible.
 
 === Machine tracking table pointer (`mttp`)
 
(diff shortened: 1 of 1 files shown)


chapter3.adoc
```diff
@@ -6,7 +6,13 @@
 under which a hart is operating. The interface consists of M-mode CSRs `msdcfg`
 and `mttp`. The SDID programmed via this interface is a local identifier for the
 hart and may be used to tag hart-local resources to access-control data
-associated with the supervisor domain.
+associated with the supervisor domain. For most typical usages, the SDID will
+remain stable over the life of the supervisor domain. However, the SDID can be
+changed at the cost of fencing to flush any prior cached state. The SDID is
+unique only within a cpu-complex, hence, SDID must not be used to tag downstream
```
ved-rivos, line 12: unique only to a hart.

rsahita, line 12:
```suggestion
unique only to a hart, hence, SDID must not be used to tag downstream
```

chapter3.adoc
```diff
@@ -6,7 +6,13 @@
...
 and `mttp`. The SDID programmed via this interface is a local identifier for the
 hart and may be used to tag hart-local resources to access-control data
-associated with the supervisor domain.
+associated with the supervisor domain. For most typical usages, the SDID will
+remain stable over the life of the supervisor domain. However, the SDID can be
+changed at the cost of fencing to flush any prior cached state. The SDID is
+unique only to a hart, hence, SDID must not be used to tag downstream
+transactions that are globally visible - implementations may map the SDID from
+the initiator to other (global) identifiers but that mapping is
+implementation-specific and out of scope of this specification.
```
SiFiveHolland, line 15: If statement 2 from #76 is correct, then we shouldn't suggest mapping the SDID to anything outside the hart:

```suggestion
transactions that are globally visible.
```

ved-rivos, line 15: We could remove the "implementations may ...." - don't need to suggest an implementation specific thing. Implementations are of course free to do whatever they need but the architecture may not want to encourage that.

SiFiveHolland (approved):

ved-rivos (approved):

Issue #62, opened 2024-09-09
Why not have fixed sized MMT tables at all levels ?


avpatel:
The MTT tables at different levels are not of same size and MMT L2 is particularly very large. This makes memory management for MTT tables more involved. Instead of this, if all MTT tables were of same size (just like MMU or IOMMU page tables) then it will be very easy to manage a pool of fixed size MMT tables.

rsahita:
this was discussed very early on in the TG - the MTT structures are biased towards minimizing lookups, and being compact. Also important is to support all arch page sizes. Another key SW aspect is top level L3, L2 MTT structures will need to be initialized early on in system initialization by the RDSM - lower level structures L1 are 4KiB based so as differing permissions are required, large pages can be fragmented easily and can L1 table pages can be flexibly allocated from pool of 4KiB pages.

rsahita:
@avpatel I added a check for pa width in the MTT walk - ptal at PR #88 
This should address your concern with overcommitting space for L3 and L2 tables.
Separately I will also send a PR for adding a type field in L3 which will allow for further flexibility.

avpatel:
Thanks, this looks reasonable for platforms with smaller system address space.

rsahita:
PR #88 addresses this issue so closing this one.

PR #88, opened 2024-10-03
add check for pa width during MTT walk

rsahita:
address issue #62 and other related

### DIFF
--- a/chapter4.adoc
+++ b/chapter4.adoc
@@ -233,15 +233,19 @@ supervisor domain are ascertained as follows:
 is 2^12^; MTT_PTE_SIZE = 8 bytes (for RV32, MTT_PTE_SIZE = 4 bytes). The `mttp`
 register must be active, i.e., the effective privilege mode must not be M-mode.
 
-2. If _pa_ has any bits set above the maximum physical address width supported
+2. If _pa_ is beyond the maximum accessible physical address space of the platform
 by the hart, then stop and raise an access-fault exception corresponding to the
 original access type.
 
 [NOTE]
 ====
-Restricting the _pa_ to the maximum PA width supported by a hart allows for
-optimizing memory requirements of the MTT structures such as the MTTL2 and MTTL3
-entry tables, when the PA width is not exactly 34, 46 or 56 bits.
+Restricting the _pa_ to the maximum accessible PA width supported by a platform allows
+for optimizing memory requirements of the MTT structures such as the MTTL2 and MTTL3
+entry tables, when the PA width is not exactly 34, 46 or 56 bits. Also note that the intent
+of using the maximum _accessible_ physical address space versus the maximum
+_implemented_ physical address width also allows for the case where an Smmtt
+mode is used that is lower width than what is implemented by the platform,
+for example, Smmtt46 used with a platform maximum PA of say 56 bits.
 ====
 
 3. Let _mpte_ be the value of the `MTT` table entry at address _a_ + _pa.pn[i]_
--- a/chapter4.adoc
+++ b/chapter4.adoc
@@ -233,26 +233,37 @@ supervisor domain are ascertained as follows:
 is 2^12^; MTT_PTE_SIZE = 8 bytes (for RV32, MTT_PTE_SIZE = 4 bytes). The `mttp`
 register must be active, i.e., the effective privilege mode must not be M-mode.
 
-2. Let _mpte_ be the value of the `MTT` table entry at address _a_ + _pa.pn[i]_
+2. If _pa_ has any bits set above the maximum physical address width supported
+by the hart, then stop and raise an access-fault exception corresponding to the
+original access type.
+
+[NOTE]
+====
+Restricting the _pa_ to the maximum PA width supported by a hart allows for
+optimizing memory requirements of the MTT structures such as the MTTL2 and MTTL3
+entry tables, when the PA width is not exactly 34, 46 or 56 bits.
+====
+
+3. Let _mpte_ be the value of the `MTT` table entry at address _a_ + _pa.pn[i]_
  x MTT_PTE_SIZE. If accessing _mpte_ violates a PMA or PMP check, raise
 an access-fault exception corresponding to the original access type.
 
-3. If any bits or encodings that are reserved for future standard use are
+4. If any bits or encodings that are reserved for future standard use are
 set within _mpte_, stop and raise an access-fault exception corresponding to
 the original access type.
 
-4. Otherwise, the _mpte_ is valid. If (_i_=1) or (_i_=2 and _mpte.type_ is not
+5. Otherwise, the _mpte_ is valid. If (_i_=1) or (_i_=2 and _mpte.type_ is not
 `MTT_L1_DIR`), go to step 5. Otherwise, the _mpte_ is a pointer to the next
 level of the `MTT`. Let _i_ = _i_-1. Let _a_ = _mpte.ppn_ x PAGESIZE and go to
 step 2. Note that when _mpte.type_ = `MTT_L1_DIR`, the _mpte.ppn_ is the value
 of the _mpte.info_ field.
 
-5. A leaf _mpte_ has been found. If any bits or encodings within _mpte.type_
+6. A leaf _mpte_ has been found. If any bits or encodings within _mpte.type_
 and _mpte.info_ that are reserved for future standard use, per
 <<Smmtt-rw-l2-encoding>>, are set within _mpte_, stop and raise an access-fault
 exception corresponding to the access type.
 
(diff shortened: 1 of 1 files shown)


chapter4.adoc
```diff
@@ -233,26 +233,37 @@ supervisor domain are ascertained as follows:
 is 2^12^; MTT_PTE_SIZE = 8 bytes (for RV32, MTT_PTE_SIZE = 4 bytes). The `mttp`
 register must be active, i.e., the effective privilege mode must not be M-mode.
 
-2. Let _mpte_ be the value of the `MTT` table entry at address _a_ + _pa.pn[i]_
+2. If _pa_ has any bits set above the maximum physical address width supported
```
ved-rivos, line 236:
```suggestion
2. If _pa_ is beyond the physical address space addressable
```

ved-rivos: The _pa_ width of a hart in RV64 is 64-bits. I think the constraint should be on the maximum accessible physical address space of the platform.

chapter4.adoc
```diff
@@ -233,26 +233,37 @@ supervisor domain are ascertained as follows:
...
+by the hart, then stop and raise an access-fault exception corresponding to the
+original access type.
+
+[NOTE]
+====
+Restricting the _pa_ to the maximum PA width supported by a hart allows for
+optimizing memory requirements of the MTT structures such as the MTTL2 and MTTL3
+entry tables, when the PA width is not exactly 34, 46 or 56 bits.
```
ved-rivos, line 244: Same comment as on the normative text to reword.

chapter4.adoc
SiFiveHolland: This also needs to handle the case where the maximum physical address width of the hart (or platform) is greater than 46, but the hart is in Smmtt46 mode.

ved-rivos: That was the intent of suggesting maximum accessible physical address space and not the implemented physical address width. That would cover that case of Smmtt mode being more restrictive but that should be called out explicitly.

chapter4.adoc
rsahita, line 244:
```suggestion
Restricting the _pa_ to the maximum accessible PA width supported by a platform allows
for optimizing memory requirements of the MTT structures such as the MTTL2 and MTTL3
entry tables, when the PA width is not exactly 34, 46 or 56 bits. Also note that the intent
of using the maximum _accessible_ physical address space versus the maximum
_implemented_ physical address width also allows for the case where an Smmtt
mode is used that is lower width than what is implemented by the platform,
for example, Smmtt46 used with a platform maximum PA of say 56 bits.
```

chapter4.adoc
```diff
@@ -233,26 +233,41 @@ supervisor domain are ascertained as follows:
...
+
+[NOTE]
+====
+Restricting the _pa_ to the maximum accessible PA width supported by a platform allows
+for optimizing memory requirements of the MTT structures such as the MTTL2 and MTTL3
+entry tables, when the PA width is not exactly 34, 46 or 56 bits. Also note that the intent
+of using the maximum _accessible_ physical address space versus the maximum
+_implemented_ physical address width also allows for the case where an Smmtt
+mode is used that is lower width than what is implemented by the platform,
+for example, Smmtt46 used with a platform maximum PA of say 56 bits.
```
SiFiveHolland, line 248: Including in this paragraph access restrictions defined by this standard is circular logic: accesses to addresses above 2^46 are disallowed with Smmtt46 because they are not "accessible", but they are only inaccessible because they are disallowed by this same paragraph. (It's not circular for some platform-specific hardware limit, since those accesses would raise an access fault anyway without Smmtt.)

Also, being vague about what makes an address "accessible" creates ambiguity: addresses blocked by PMPs aren't accessible, so can I use a smaller MTT if I restrict S-mode to a smaller portion of the address space with PMPs?

SiFiveHolland (changes requested):

chapter4.adoc
rsahita: Agree with the confusing "accessible" term - updating that to use "maximum-addressable" instead.

Re. the implementation note - so PMP are  the last thing to be checked (and hence MTT walks are subject to PMPs), but both MTT and PMP have to allow the access for an address to become accessible - so if the PMP  settings on a platform blocks access to a smaller region of memory than available on a platform, then yes, the MTT would only need to cover that region, since any access that violates that would be blocked by PMPs.

will update the implementation note as well to use "addressable" vs "accessible" - and the second part is probably confusing and not needed - so only keeping the software optimization of memory use for MTT structures in the note.

SiFiveHolland (approved):
LGTM

chapter4.adoc
SiFiveHolland: The two statements

> so PMP are the last thing to be checked (and hence MTT walks are subject to PMPs)

and

> yes, the MTT would only need to cover that region, since any access that violates that would be blocked by PMPs.

seem to be contradictory to me. If the PMPs are checked last, then the hardware could have started the MTT table walk, and thus performed an out-of-bounds read (possibly chasing a wild pointer from an out-of-bounds MTTL2 entry to an arbitrary "MTTL1" address) before the PMP rejects the original access.

ved-rivos: The hardware is not allowed to bypass PMP/PMA for MTT walks. MTT accesses are themselves subject to PMA/PMP rules. Whether an implementation checks PMP for the original access in parallel with doing the MTT walk or does it serially (either before or after MTT walk) is implementation choice - the result should be the same i.e. original memory access is allowed only if PMP/PMA/MTT allow the access.

SiFiveHolland: > The hardware is not allowed to bypass PMP/PMA for MTT walks. MTT accesses are themselves subject to PMA/PMP rules.

"MTT structure accesses are to be treated as implicit M-mode accesses", so they use a different set of PMP permissions than the original S/U-mode access. And the MTT structure accesses are at a totally different address than the original access.

Here's an example of what I mean:

Say I have a 32-bit system where the physical address space contains 1G of MMIO (0x00000000-0x40000000) followed by 2G of RAM (0x40000000-0xc0000000). So the platform's PAW is 32 bits. I then configure the PMP rules to restrict S-mode to the address range 0x00000000-0x80000000. You guys are saying I only need an MTTL2 with 64 entries, large enough to cover 31 bits of physical address space. I allocate the MTTL2 at, say, address 0x40100000. The memory immediately following the MTTL2 table (at 0x40100200) happens to contain other arbitrary data, with the first 4 bytes coincidentally having the value 0x01010000.

Now my S-mode software tries to access the address 0x80000004. My hardware does the MTT walk before the PMP check. The MTT walk accesses the out-of-bounds MTTL2 entry at 0x40100200 (because the address passes the PAW bounds check). This access passes the PMP check, because 1) the address 0x40100200 is in the allowed PMP range, and 2) the MTT walk operates with M-mode permissions anyway. This is a valid MTTL2 entry, pointing to an MTTL1 at physical address 0x10000000.

The MTT walk then accesses the first entry of (what it thinks is an) MTTL1 table at 0x10000000. This access also passes the PMP check, for the same two reasons. But 0x10000000 is really the MMIO address of my NS16550 UART, so the MTT checker has just consumed and discarded a character from the UART's receive FIFO. The character happens to be "w" (0x77), which has the low two bits set, so the MTT checker interprets this as allowing read, write, and execute access.

Now the PMP check for the original S-mode access is performed. It denies the access, because 0x80000004 is outside the range allowed by the PMP rules, but the MTT checker has already eaten a byte of my UART input! This is clearly a different result than if the PMP rejection of the S-mode access prevented the MTT walk from occurring entirely.

rsahita: A clarifying question - if an implementation follows the spec with the updated step 2 - where the pa should not be beyond the platform-defined maximum-addressable physical address for the hart (which should be 0x80000000 in your example), then the MTT walk should not access the out-of-bounds MTTL2 entry at 0x40100200?

SiFiveHolland: No, the platform-defined maximum-addressable physical address for the hart must be at least 0xbfffffff, because that's the end of DRAM. The addresses from 0x80000000-0xc0000000 are _still currently accessible_ by M-mode on that hart even with the PMP configuration I described. More generally, I would consider an address range to be addressable even if access is being temporarily blocked by a permission check.

Is your interpretation that any memory access that, in the absence of Smmtt, would raise an access fault for absolutely any reason, should fail at step 2? If so, then we should just say that. But that wasn't my interpretation, and I don't know if that's reasonable from a hardware design perspective.

rsahita: sorry yes I misread - so in your example, the MTT L2 would need to cover the 3G addressable memory space.

SiFiveHolland: Got it, thanks!

ved-rivos (approved):

PR #93, opened 2024-10-10
specify a reset for SDICN

ved-rivos:


chapter7.adoc
```diff
@@ -160,6 +160,9 @@ domains. To emulate an IMSIC, the RDSM may use the illegal instruction trap on
 access to the associated CSRs for emulation purposes.
 ====
 
+If `Smsdia` extension is implemented, then `msdcfg.SDICN` is non-zero and holds
+the number of an implemented supervisor domain interrupt controller.
```
SiFiveHolland, line 164: Maybe I'm missing something, but I don't see anything to imply this sentence specifies a value at reset. Should there be a heading or something like the following?

```suggestion
If `Smsdia` extension is implemented, then at reset `msdcfg.SDICN` is non-zero and holds
the number of an implemented supervisor domain interrupt controller.
```

ved-rivos: Yes, its missing that qualification.

SiFiveHolland (approved):

rsahita (approved):

Issue #106, opened 2024-10-28
Level of MTT when mode is Smmtt46


sqzsq:
This could be possibly a typo, but still:

Section 4.1 says:

"Figure 7 shows an RV64 supervisor physical address lookup; for lower physical address widths e.g. 34 or 46 bits, the **MTTL3 table is not applicable** as shown in Figure 8." 

while Section 4.2 says:

"1. Let a be mttp.ppn x PAGESIZE, and let i = LEVELS, where for mode Smmtt34, LEVELS = 2 and for **Smmtt[46 | 56], LEVELS = 3**;"

rsahita:
yup typo - should be LEVELS = 2 for Smmpt[34 | 46] and LEVELS = 3 for Smmpt56

rsahita:
addressed in PR #107

PR #107, opened 2024-10-29
address issue 106

rsahita:
address issue #106


rsahita:
minor fix so merging without additional review.

PR #112, opened 2024-11-15
Correct top-level MPT size/alignment requirements

### LINKED ISSUE #111: Should the mptl3.ppn be aligned with the mpt table size (16MB)?
**Should the mptL3 ppn  be aligned with the mpt table size?**

for MPTL3, the mptl3.ppn is the base addr for mptL2, which is size of 16MB, 

should the  mptl3.ppn be 

A : aligned with the mpt table size(16MB)?

B: or any addr without any alignment restriction?

SiFiveHolland:
Closes #111

* The first sentence is wrong because the type of table depends on the mode (even if the PAW is larger or smaller).
* The sizes given are wrong for Smmpt34, because in that case each entry is 4 bytes, not 8.
* The minimum alignment (irrespective of table size) is 4 KiB because that is the number of low bits omitted from the PPN.

### DIFF
--- a/chapter3.adoc
+++ b/chapter3.adoc
@@ -112,13 +112,19 @@ Implementations are not required to support all defined `MODE` settings when
 Instead, the fields of `mmpt` are `WARL` in the normal way, when so indicated
 in <<sdid-32>> and <<sdid-64>>.
 
-The `mmpt.PPN` refers to an `MPTL3` table or an `MPTL2` table based on physical
-address width (`PAW`). For 56 >= `PAW` > 46, `MPTL3` table must be of size
-`2^(PAW-43)` bytes and naturally aligned to that sized byte boundary. For 46
->= `PAW` > 32 the `MPTL2` table must be of size 2^(`PAW`-22) bytes for
-`Smmpt46` and `Smmpt34`, and must be naturally aligned to that sized byte
-boundary. In these modes, the lowest two bits of the physical page number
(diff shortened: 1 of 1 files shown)


chapter3.adoc
```diff
@@ -112,13 +112,19 @@ Implementations are not required to support all defined `MODE` settings when
...
+physical address width (`PAW`), which is the number of bits needed to express
+the platform-defined maximum-addressable physical address.
+
+For `PAW` >= 56 a top-level `MPTL3` table must be complete; for 56 > `PAW` >=
+46 the table must contain at least `2^(PAW-46)` entries; otherwise it must
+contain at least one entry. For `PAW` >= 46 (34 for `Smmpt34`) a top-level
+`MTTL2` table must be complete; otherwise for `PAW` >= 25 the table must
+contain at least `2^(PAW-25)` entries; otherwise it must contain at least one
+entry. The top-level table must be naturally aligned to its size, or aligned to
+a 4 KiB page boundary, whichever is greater.
```
rsahita, line 127: why the "must contain at least one entry"? @SiFiveHolland

SiFiveHolland: Because you can use Smmpt56 even if you have <46 bits of PAW.

rsahita: makes sense - thanks

rsahita (approved):
lgtm

PR #121, opened 2025-02-04
Updated Smmpt to address ARC feedback

rsahita:
Support upto 64 bit address (52 bit ppn at each non-leaf)
Non-leaf entry has ppn, valid bit + napot bit
Leaf entry updated with RWX perms
Napot for non-leaf allows entries to be coalesced (larger next level)
Napot in leaf supports super page sizes
Common non-leaf and leaf format

See this presentation for an overview (the format has been updated from the presentation so refer to it only as an overview - read specifics in the PR please): https://docs.google.com/presentation/d/1Zzd5Jm6HeITo4vjJzoAnOp9wGHsOVPEI8FL2BJc8YsE/edit#slide=id.g32322424817_0_7


chapter4.adoc
```diff
@@ -2,8 +2,8 @@
 [[Smmpt]]
 == `Smmpt`: Hart MPT Extension
 
-[[Smmpt-rw]]
-=== Smmpt[34 | 46 | 56]
+[[Smmpt-rwx]]
+=== Smmpt
 
 The Supervisor Physical Address (SPA) for RV32 and RV64 is shown below.
```
ved-rivos, line 8:
```suggestion
The Supervisor Physical Address (SPA) for RV32 and RV64 are partitioned into a Page Number (pn)  and a page offset as shown in <<add figure reference>> and <<add figure reference>> respectively.
```

rsahita:
[smmtt-spec-PR-121-2-16-2025.pdf](https://github.com/user-attachments/files/18816297/smmtt-spec-PR-121-2-16-2025.pdf)

PDF to make reviewing easier.

ved-rivos (approved):

chapter4.adoc
```diff
@@ -1,390 +1,426 @@
...
+  {bits:  1,  name: 'N'},
+], config:{lanes: 2, hspace:1024}}
+....
 
-[caption="Figure {counter:image}: ", reftext="Figure {image}"]
-[title= "`MPT` walk for RV32", id=mpt-map-rv32]
-image::images/fig3b.png[]
+The `V` bit indicates if the MPTE is valid; if it is 0, all other bits in the
+MPTE are don't-cares and may be used freely by software. When the `MPTE` is
+valid, the `RESERVED` field must be 0.
```
ved-rivos, line 78:
```suggestion
valid, the bits constituting the `RESERVED` field are reserved for future standard use and, until its use is defined by some standard extension, must be zeroed by software for forward compatibility. If any of these bits are set, an access-fault exception is raised.
```

chapter4.adoc
```diff
@@ -1,390 +1,426 @@
...
+  {bits:  1,  name: 'N'},
+], config:{lanes: 2, hspace:1024}}
+....
 
-[caption="Figure {counter:image}: ", reftext="Figure {image}"]
-[title= "`MPT` walk for RV32", id=mpt-map-rv32]
-image::images/fig3b.png[]
+The `V` bit indicates if the MPTE is valid; if it is 0, all other bits in the
+MPTE are don't-cares and may be used freely by software. When the `MPTE` is
+valid, the bits constituting the `RESERVED` field are reserved for future standard use and, until its use is defined by some standard extension, must be zeroed by software for forward compatibility. If any of these bits are set, an access-fault exception is raised.
```
rsahita, line 78:
```suggestion
valid, the bits constituting the `RESERVED` field are reserved for future
standard use and, until its use is defined by some standard extension,
must be zeroed by software for forward compatibility. If any of these
bits are set, an access-fault exception is raised.
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+
+Access type permissions -- readable, writable, or executble -- are checked
+by MPT the same as for VS-stage and G-stage translation. For a memory access
+made to support VS-stage or G-stage address translation (such as a read/write to
+a VS-level page table or a G-level page table), permissions are checked as
+though for an implicit load or store, not for the original access type. However,
+any exception is always reported for the original access type (instruction,
+load, or store/AMO).
+
+If Smmptnapot is not implemented, the bit 31 in leaf MPTE remains reserved.
```
SiFiveHolland, line 128: editorial:

```suggestion
If Smmptnapot is not implemented, bit 31 in a leaf MPTE remains reserved.
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+4. Otherwise, the _mpte_ is valid. If _mpte.L=1_ go to step 5; Otherwise, this
+   MPTE is a pointer to the next level of the memory protection table. Let
+   _i_=_i_-1. If _i_<0, stop and raise an access-fault exception corresponding
+   to the original access type. Otherwise, let _a_ = _mpte.PPN_ x PAGESIZE and
+   go to step 2.
+
+5. A leaf _mpte_ has been found. Let _pi_ be the NUMPGINRANGE (For Smmpt34,
+   NUMPGINRANGE is 3) most significant bits of the range offset. Let
+   _XWR=PERMS[pi*3+2:pi*3]_. Determine if the requested memory access is allowed
+   by the _XWR_ bits, given the current privilege mode and MXR field of the
```
SiFiveHolland, line 160:
```suggestion
   by the _XWR_ bits, given the effective privilege mode and MXR field of the
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+   corresponding to the original access type.
+
+4. Otherwise, the _mpte_ is valid. If _mpte.L=1_ go to step 5; Otherwise, this
+   MPTE is a pointer to the next level of the memory protection table. Let
+   _i_=_i_-1. If _i_<0, stop and raise an access-fault exception corresponding
+   to the original access type. Otherwise, let _a_ = _mpte.PPN_ x PAGESIZE and
+   go to step 2.
+
+5. A leaf _mpte_ has been found. Let _pi_ be the NUMPGINRANGE (For Smmpt34,
+   NUMPGINRANGE is 3) most significant bits of the range offset. Let
```
SiFiveHolland, line 158: For i > 0, _pi_ is the NUMPGINRANGE most significant bits of _pa.pn[i-1]_, _not_ the bits from the range offset.

Also, I would recommend renaming this to something like `PGBITSINRANGE`, since it is really the log base 2 of the number of pages in the range.

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+   corresponding to the original access type, else the requested memory access
+   is allowed.
+
+Implicit accesses to MPT by this process are checked by PMP/Smepmp as implicit
+M-mode accesses. The endianness of such implicit accesses is controlled by
+`mstatus.MBE`.
+
+=== Smmpt43: Page-Based 43-bit Memory Protection System
+
+This section describes a page-based memory protection system RV64, which
```
SiFiveHolland, line 171: editorial:

```suggestion
This section describes a page-based memory protection system for RV64, which
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+supports a 43-bit supervisor physical address space. The design of Smmpt43
+follows the overall scheme of Smmpt34, and this section details only the
+differences between the schemes.
+
+[NOTE]
+====
+Multiple memory protection systems are specified for RV64 to relieve the tension
+between providing a large physical address space and minimizing the memory
+protection cost. For many systems, 43-bits of physical address space is ample,
```
SiFiveHolland, line 180: For many RV64 systems, 34 bits of physical address space is also ample. We should define a two-level mode for RV64 as well. Unfortunately, the naming could be confusing, as it conflicts with the RV32 Smmpt34.

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
 [wavedrom, ,svg]
 ....
 {reg: [
-  {bits:  44, name: 'MPTL2 PPN'},
-  {bits:   1, name: 'VALID'},
-  {bits:  19, name: 'RESERVED'},
+  {bits:  16, name: 'range offset', attr: '16'},
+  {bits:  9,  name: 'pn[0]',        attr: '9' },
+  {bits:  9,  name: 'pn[1]',        attr: '9'},
+  {bits:  9,  name: 'pn[2]',        attr: '9'},
```
SiFiveHolland, line 201: The diagram should include all 64 bits, with the upper bits set to zero per the text above.

```suggestion
  {bits:  9,  name: 'pn[2]',        attr: '9'},
  {bits:  21, name: '0',            attr: '21'},
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
-_The 32 MiB range of address space is partitioned into 16 2 MiB pages where each
-page has read/write/execute access specified via the `INFO` field._
-The `INFO` field holds 16 2-bit `PERM` access-encoding for each 2 MiB address
-range to indicate: `no_access_allowed` (00b), `read-execute-allowed` (01b),
-`read-write-allowed` (10b), `read-write-execute-allowed` (11b). Bits
-_mpte.info[31:0]_ holds the 2 MiB page permissions, and _mpte.info[43:32]_ are
-reserved and must be zero.
-|===
+If Smmptnapot is not implemented, the bit 63 in leaf MPTE remains reserved. If
+Smmptnlnapot is not implemented, the bit 63 in non-leaf MPTE remains reserved.
```
SiFiveHolland, line 246: editorial:

```suggestion
If Smmptnapot is not implemented, bit 63 in a leaf MPTE remains reserved. If
Smmptnlnapot is not implemented, bit 63 in a non-leaf MPTE remains reserved.
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
 {reg: [
-  {bits:  2,  name: 'PERM'},
-  {bits:  28, name: '...'},
-  {bits:  2,  name: 'PERM'},
-  {bits:  32, name: 'resv'},
+  {bits:  16, name: 'range offset', attr: '16'},
+  {bits:  9,  name: 'pn[0]',        attr: '9' },
+  {bits:  9,  name: 'pn[1]',        attr: '9'},
+  {bits:  9,  name: 'pn[2]',        attr: '9'},
+  {bits:  9,  name: 'pn[3]',        attr: '9'},
```
SiFiveHolland, line 278: Likewise this diagram should include all 64 bits of the PA:

```suggestion
  {bits:  9,  name: 'pn[3]',        attr: '9'},
  {bits:  12, name: '0',            attr: '12'},
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
 ], config:{lanes: 1, hspace:1024}}
 ....
 
+The Smmpt52 memory protection tables contain 2^9^ memory protection table
+entries (MPTEs). Each MPTE is an eight byte value. A page table is exactly the
+size of a page and must always be aligned to a page boundary. The physical page
+number of the root memory protection table is stored in the `mmpt` register's
+PPN field.
+
+The MPTE formats for Smmpt52 is identical to that of Smmpt43.
```
SiFiveHolland, line 288: editorial:

```suggestion
The MPTE formats for Smmpt52 are identical to that of Smmpt43.
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+=== Smmpt64: Page-Based 64-bit Memory Protection System
+
+This section describes a page-based memory protection system for RV64 that
+supports 64-bit physical address spaces. It closely follows the design of
+Smmpt52, simply adding an additional level of page table, and so this section
+only details the differences between the two schemes.
+
+Smmpt64 supports a 64-bit physical address space, divided into pages. The
+Smmpt64 address is partitioned as shown in  <<rv64-spa-Smmpt64>>. The
+Smmpt64 memory protection table has four levels.
```
SiFiveHolland, line 307:
```suggestion
Smmpt64 memory protection table has five levels.
```

chapter3.adoc
```diff
@@ -52,137 +50,75 @@ an illegal instruction exception.
...
-reused, it may be necessary to execute a `MFENCE.SPA` instruction before or
-after writing `mmpt`.
+Implementations are not required to support all defined `MODE` settings. A write
+to `mmpt` with an unsupported `MODE` value is not ignored. Instead, the fields
+of `mmpt` are `WARL` in the normal way, when so indicated in <<sdid-32>> and
+<<sdid-64>>.
 
-=== Machine supervisor domain configuration (`msdcfg`)
+`mmpt.PPN` refers to the root of memory protection table when `mmpt.MODE` is not
+`Bare`.
```
SiFiveHolland, line 107:
```suggestion
`Bare`. When `mmpt.MODE` is `Smmpt34`, `mmpt.PPN`[0] must be zero. When
`mmpt.MODE` is `Smmpt64`, `mmpt.PPN`[2:0] must be zero.
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
-  {bits:  9,  name: 'pn[2]',       attr: '9'},
+  {bits:  15, name: 'range offset', attr: '15'},
+  {bits:  10, name: 'pn[0]',        attr: '10' },
+  {bits:  9,  name: 'pn[1]',        attr: '9'},
 ], config:{lanes: 1, hspace:1024}}
 ....
 
+The root memory protection table has 2^9^ memory-protection-table entries
+(MPTEs) and the memory protection tables linked to the root table have 2^10^
+MPTEs. Each MPTE is a four byte value.
```
SiFiveHolland, line 42:
```suggestion
MPTEs. Each MPTE is a four byte value. The root memory protection table
is 8 KiB instead of the usual 4 KiB and must be aligned to an 8 KiB boundary.
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
-supervisor domain are ascertained as follows:
+The non-root Smmpt64 memory protection tables contain 2^9^ MPTEs, eight bytes
+each. The non-root memory protection table is exactly the size of a page and
+must always be aligned to a page boundary.
 
-1. Let _a_ be `mmpt.ppn` x PAGESIZE, and let _i_ = LEVELS, where for mode
-`Smmpt[34 | 46]`, LEVELS = 2 and for `Smmpt56`, LEVELS = 3; PAGESIZE
-is 2^12^; MPT_PTE_SIZE = 8 bytes (for RV32, MPT_PTE_SIZE = 4 bytes). The `mmpt`
-register must be active, i.e., the effective privilege mode must not be M-mode.
+The MPTE formats for Smmpt64 is identical to that of Smmpt52.
```
SiFiveHolland, line 333: editorial:

```suggestion
The MPTE formats for Smmpt64 are identical to that of Smmpt52.
```

SiFiveHolland (commented):
Here is my review of everything except the NAPOT sections; I will submit that separately due to GitHub lag.

chapter3.adoc
```diff
@@ -52,137 +50,75 @@ an illegal instruction exception.
...
-
-Note that writing `mmpt` does not imply any ordering constraints between
-`S-mode` and `G-stage` page-table updates and subsequent address translations.
-If a supervisor domain's `MPT` structure has been modified, or if a `SDID` is
-reused, it may be necessary to execute a `MFENCE.SPA` instruction before or
-after writing `mmpt`.
+Implementations are not required to support all defined `MODE` settings. A write
+to `mmpt` with an unsupported `MODE` value is not ignored. Instead, the fields
+of `mmpt` are `WARL` in the normal way, when so indicated in <<sdid-32>> and
+<<sdid-64>>.
```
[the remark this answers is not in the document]
ved-rivos: This was left out on purpose. In case of Svxx extensions, one could boot an older operating system that only used Sv39 - for instance in a virtual machine.  Setting up a mmpt mode smaller than the physical memory in the machine does not help much as it would only lead to stranded memory. For the mmpt usage, the machine mode needs to be matched to the machine.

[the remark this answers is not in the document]
SiFiveHolland: Using the smallest possible Smmpt mode is helpful because it increases the performance of MPT walks. The mode can be chosen based on the range of addresses allocated to a given supervisor domain, regardless of the total amount of physical memory in the system.

M-mode firmware can automatically choose the smallest mode for each supervisor domain, and adjust the mode as memory allocations change. This can be done efficiently because addresses are zero-extended, so switching between modes doesn't require rewriting any of the tables.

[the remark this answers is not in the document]
ved-rivos: The mmpt - indexed by the SPA; unlike VA - so to use a smaller mode, the memory for that domain would also need to be located in lower memory..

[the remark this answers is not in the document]
SiFiveHolland: Yes, that is true. And the RDSM or system designer can take that into account when partitioning memory between supervisor domains, by preferring to allocate memory for smaller supervisor domains (and shared memory) from the bottom of the physical address space.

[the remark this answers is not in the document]
ved-rivos: Besides NL napot provide a way to short circuit the walk using a single cached PTE entry . So while a machine can choose to provide all smaller modes it does not need to be mandated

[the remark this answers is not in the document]
ved-rivos: Even without NL napot if the upper levels have a single valid entry because that’s all the memory mapped then just a single PTE entry cache suffices to short circuit the walk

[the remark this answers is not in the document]
SiFiveHolland: NAPOT entries solve an entirely different problem. They make large allocations more efficient. I'm talking about making small allocations more efficient when the high bits are known to be zero.

Also consider the case where a system has socketed memory or an otherwise variable amount of memory -- the CPU may support a larger physical address space (and thus a larger Smmpt mode), but M-mode firmware knows at runtime that less physical address space is actually present/usable. So even if M-mode chooses a single mode for all supervisor domains, based on e.g. the devicetree /memory node, it may still not choose the largest MPT mode supported by the CPU.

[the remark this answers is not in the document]
ved-rivos: Since the high bits need to be 0 and it’s a radix tree - one PTE entey needs to be cached at each level to short the walk - the entry 0. A machine built for N levels would have PTE caches for each level . If a level uses only entry 0 as valid it only underutilizwa what that machine is built for

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
-and _mpte.info_ that are reserved for future standard use, per
-<<Smmpt-rw-l2-encoding>>, are set within _mpte_, stop and raise an access-fault
-exception corresponding to the access type.
+Such MPTEs behave identically to leaf MPTEs in the MPT access permission lookup
+process described in <<MPT-lookup>>, except that:
 
-7. The _mpte_ is a valid leaf _mpte_. Fetch the access-permissions for the
-physical address per the steps described below:
+* For RV32, the `N` bit may be set to 1 in the level 0 leaf MPTEs.
+* For RV64, the `N` bit may be set to 1 in the level 0 and level 1 leaf MPTEs.
```
SiFiveHolland, line 356: Why are leaf NAPOT MPTEs not allowed at all levels? Or all non-root levels? Sv48 and Sv57 have larger leaf page sizes.

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
-`4M_PAGES`, the _mpte.info_[16:0] field contains 8 entries of 2-bit
-access-permission encodings for 8 4 MiB regions - see <<Smmpt-rw-l2-encoding>>;
-go to step 8, else
+[NOTE]
+====
+The motivation of NAPOT MPTE is that access-type permissions for one or more
+entries representing the contigous region may be cached as a single cache entry
+as if it were a single (large) page. This compaction can relieve access-type
+permission caches. The scheme allows an implementation to not take advantage of
+this property and simply cache the access-type permissions for pages seperately.
```
SiFiveHolland, line 364: This is a good point. Implementations can completely ignore the N bit if they want to. So there's no hardware cost for implementing Smmptnapot. I don't think it should be a separate extension. Svnapot had to be separate because the N bit in PTEs was previously reserved. But since we are defining the N bit in MPTEs from the start, there are no existing Smmpt implementations to be confused by it.

chapter3.adoc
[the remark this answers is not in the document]
SiFiveHolland: OK, that's a reasonable argument. Hardware large enough to support more than the smallest MPT mode is likely large enough to have a MPTE cache.

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+this NAPOT memory protection range have the same value for L, N, PERMS and V
+bits. Further, the XWR access type permssion tuples in the PERMS field of such
+PTEs are identical.
 
-6. A leaf _mpte_ has been found. If any bits or encodings within _mpte.type_
-and _mpte.info_ that are reserved for future standard use, per
-<<Smmpt-rw-l2-encoding>>, are set within _mpte_, stop and raise an access-fault
-exception corresponding to the access type.
+Such MPTEs behave identically to leaf MPTEs in the MPT access permission lookup
+process described in <<MPT-lookup>>, except that:
```
SiFiveHolland, line 353: broken link:

```suggestion
process described in <<MPT_ACC_LKUP>>, except that:
```

SiFiveHolland (commented):
I've provided my comments on Smmptnlnapot as a pull request (#126) because I mostly rewrote the section to make it (hopefully) easier to understand.

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+=== Smmptnapot - NAPOT Access-Type Permission Granularity
 
-5. Otherwise, the _mpte_ is valid. If (_i_=1) or (_i_=2 and _mpte.type_ is not
-`MPT_L1_DIR`), go to step 6. Otherwise, the _mpte_ is a pointer to the next
-level of the `MPT`. Let _i_ = _i_-1. Let _a_ = _mpte.ppn_ x PAGESIZE and go to
-step 3. Note that when _mpte.type_ = `MPT_L1_DIR`, the _mpte.ppn_ is the value
-of the _mpte.info_ field.
+When `MPTE.L`=1 and `MPTE.N`=1, the leaf `MPTE` represents a memory protection
+range that is part of a larger contigous NAPOT memory protection range comprised
+of 512 pages for RV64 and 1024 pages for RV32. All leaf MPTEs at that level of
```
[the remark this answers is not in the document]
ved-rivos: The merging in TLB is not disallowed when vs and g stage resolve to 64K and the mpt provides a page size of > 4K.

merging becomes disallowed when the effective page size determined by Vs and G stage is smaller than page size in mpt and not vice versa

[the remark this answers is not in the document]
SiFiveHolland: It also becomes a problem if the MPTE size is smaller than the VS/G-stage page size, because then the PTW would have to read multiple MPTEs to determine if the entire page can be accessed. But now I realize this means Svnapot's 64K page size will work fine, because one MPTE covers exactly 64K. The hardware will just need logic to compare all of the PERMS entries within the MPTE and split the page if they are inconsistent. So it's only a concern for new NAPOT page sizes larger than 64K, and maybe that's not worth adding complexity to support.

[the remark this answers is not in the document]
ved-rivos: Its not a problem when MPTE is smaller - the permissions are checked only for the bytes that are access. Consider a VS stage PTE creating a 1G page and the G stage mapping that 1G region using 256K 4K pages. This does not prevent the VS stage from accessing any of the 1G page. However, usually its not cached as a 1G page in the TLBs since caching as a 1G translation would require caching 256K SPA+G-stage-permission-sets in that TLB entry.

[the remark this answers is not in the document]
SiFiveHolland: > the permissions are checked only for the bytes that are access.  

I know that, but if you want to avoid splitting the page at the TLB, you have to know the permissions for the whole page.

In your example, the TLB entry is 4K, because that's the smallest page size in the lookup path. There will be reduced performance due to increased TLB pressure, but that's a software problem: software _should_ have used 1G pages for the G-stage mapping (barring fragmentation), and then the TLB could use a 1G entry.

My claim is that when the NAPOT VM page size is larger than an MPTE, the TLB entry for that page will _necessarily_ be split, because the "smallest page size in the lookup path" will always be smaller than the NAPOT VM page size, thus negating the performance benefits of using the NAPOT VM page -- that's the problem I am referring to. If I allocate 256K to a supervisor domain, there's no way to make that one NAPOT page in the MPT, so no matter what software does, it cannot get full performance from a hypothetical 256K NAPOT VM page.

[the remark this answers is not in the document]
ved-rivos: When a 256K NAPOT page gets defined then along with such a future extension there may be an extension to Smmpt.

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
 [[chapter4]]
 [[Smmpt]]
-== `Smmpt`: Hart MPT Extension
+== `Smmpt`: Machine-mode Memory Protection Table (MPT)
```
rsahita, line 3:
```suggestion
== Machine-mode Memory Protection Table (MPT)
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
 
-[[Smmpt-rw]]
-=== Smmpt[34 | 46 | 56]
+Memory Protection table is a radix-tree table used to provide read, write, and
+execute permissions for implicit and explicit memory accesses by effective
+privilege modes less than M.
 
-The Supervisor Physical Address (SPA) for RV32 and RV64 is shown below.
-The MPT maps the SPA to read-write-execute access permissions.
+The following MPT extensions are defined:
```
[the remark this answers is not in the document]
rsahita: ack - removed

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
+The `V` bit indicates if the MPTE is valid; if it is 0, all other bits in the
+MPTE are don't-cares and may be used freely by software. When the `MPTE` is
+valid, the bits constituting the `RESERVED` field are reserved for future
+standard use and, until its use is defined by some standard extension,
+must be zeroed by software for forward compatibility. If any of these
+bits are set, an access-fault exception is raised.
+
+Explicit stores by a hart that update the `V` bit of leaf and/or non-leaf MPTEs
+from 0 to 1 and are visible to a hart will eventually become visible within a
+bounded timeframe to subsequent implicit accesses by that hart to such MPTEs.
```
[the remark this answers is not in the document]
rsahita: I can add the following to the NOTE below (modeled after the note in Svvptc), which is what we should require IMO to provide implementation flexibility:

"This behavior relieves M-mode software from executing certain memory-management
instructions, such as MFENCE.SPA or MINVAL.SPA, which would normally be used to
synchronize the hart’s MPT caches when a memory-resident MPTE is changed from
Invalid to Valid. Synchronizing the hart’s MPT caches with other forms of updates
to a memory-resident MPTE, including when a MPTE is changed from Valid to Invalid,
requires the use of suitable memory-management instructions. This behavior
guarantees that a change to a MPTE from Invalid to Valid is made visible within a
bounded time, thereby making the execution of these memory-management
instructions redundant. The performance benefit of eliding these instructions
outweighs the cost of an occasional gratuitous additional access fault that may occur."

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
-`MPT_L1_DIR`), go to step 6. Otherwise, the _mpte_ is a pointer to the next
-level of the `MPT`. Let _i_ = _i_-1. Let _a_ = _mpte.ppn_ x PAGESIZE and go to
-step 3. Note that when _mpte.type_ = `MPT_L1_DIR`, the _mpte.ppn_ is the value
-of the _mpte.info_ field.
+When `MPTE.L`=1 and `MPTE.N`=1, the leaf `MPTE` represents a memory protection
+range that is part of a larger contiguous NAPOT memory protection range comprised
+of 512 pages for RV64 and 1024 pages for RV32. All leaf MPTEs at that level of
+this NAPOT memory protection range have the same value for L, N, PERMS and V
+bits. Further, the XWR access type permssion tuples in the PERMS field of such
+PTEs are identical.
```
rsahita, line 350:
```suggestion
bits. Thus, the XWR access type permission tuples in the PERMS field of such
PTEs are identical.
```

chapter4.adoc
```diff
@@ -1,390 +1,429 @@
...
-`MPT_L1_DIR`), go to step 6. Otherwise, the _mpte_ is a pointer to the next
-level of the `MPT`. Let _i_ = _i_-1. Let _a_ = _mpte.ppn_ x PAGESIZE and go to
-step 3. Note that when _mpte.type_ = `MPT_L1_DIR`, the _mpte.ppn_ is the value
-of the _mpte.info_ field.
+When `MPTE.L`=1 and `MPTE.N`=1, the leaf `MPTE` represents a memory protection
+range that is part of a larger contigous NAPOT memory protection range comprised
+of 512 pages for RV64 and 1024 pages for RV32. All leaf MPTEs at that level of
+this NAPOT memory protection range have the same value for L, N, PERMS and V
+bits. Further, the XWR access type permssion tuples in the PERMS field of such
+PTEs are identical.
```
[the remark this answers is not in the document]
rsahita: I replaced "Further" with "Thus" - good to call this out explicitly.

chapter4.adoc
rsahita: I think we cover all the cases expected currently? other Napot encodings can be added in the future.

chapter4.adoc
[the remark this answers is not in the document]
SiFiveHolland: Yes, I think that would be helpful, and sufficient to resolve my concern. That at least makes it explicit that gratuitous access faults may still occur, which we would expect if we are following the behavior of Svvptc.

chapter4.adoc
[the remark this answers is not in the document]
SiFiveHolland: OK, that makes sense. We would use some of the reserved bits to define a new NAPOT MPT page size if a new VM page size is defined.

chapter4.adoc
rsahita: i think it is still good to keep this modular - this aspect is being discussed in the other PR you submitted so will resolve this comment.

chapter3.adoc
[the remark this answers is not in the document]
rsahita: add to the existing non-normative NOTE in the SmmptNapot section-  a single MPTE entry covers 64K address range, so a seperate 64KiB napot is not provided. All perms identical in an MPTE provides this support.

chapter3.adoc
rsahita: we already say in 4.4: The root memory protection table is 32 KiB instead of the usual 4
KiB and must be aligned to a 32 KiB boundary. The physical page number of the root memory
protection table is stored in the mmpt register’s PPN field.

chapter4.adoc
```diff
@@ -1,390 +1,432 @@
...
+from 0 to 1 and are visible to a hart will eventually become visible within a
+bounded timeframe to subsequent implicit accesses by that hart to such MPTEs.
+
+[NOTE]
+====
+Depending on the microarchitecture, some possible ways to facilitate include:
+not storing Invalid PTEs in the MPT page-walk caches, automatically evicting
+Invalid MPTEs using a bounded timer, or making MPT page-walk caches coherent
+with store instructions that modify MPTEs.
+====
```
rsahita, line 94:
```suggestion
[NOTE]
====
This behavior relieves M-mode software from executing certain memory-management
instructions, such as MFENCE.SPA or MINVAL.SPA, which would normally be used to
synchronize the hart’s MPT caches when a memory-resident MPTE is changed from
Invalid to Valid. Synchronizing the hart’s MPT caches with other forms of updates
to a memory-resident MPTE, including when a MPTE is changed from Valid to Invalid,
requires the use of suitable memory-management instructions. This behavior
guarantees that a change to a MPTE from Invalid to Valid is made visible within a
bounded time, thereby making the execution of these memory-management
instructions redundant. The performance benefit of eliding these instructions
outweighs the cost of an occasional gratuitous additional access fault that may occur.

Depending on the microarchitecture, some possible ways to facilitate include:
not storing Invalid PTEs in the MPT page-walk caches, automatically evicting
Invalid MPTEs using a bounded timer, or making MPT page-walk caches coherent
with store instructions that modify MPTEs.
====
```

chapter4.adoc
```diff
@@ -1,390 +1,432 @@
...
 ....
 {reg: [
-  {bits:  44, name: 'MPTL2 PPN'},
-  {bits:   1, name: 'VALID'},
-  {bits:  19, name: 'RESERVED'},
+  {bits:  16, name: 'range offset', attr: '16'},
+  {bits:  9,  name: 'pn[0]',        attr: '9' },
+  {bits:  9,  name: 'pn[1]',        attr: '9'},
+  {bits:  9,  name: 'pn[2]',        attr: '9'},
+  {bits:  21, name: '0',            attr: '21'},
```
rsahita, line 203:
```suggestion
  {bits:  16, name: 'range offset', attr: '16'},
  {bits:  9,  name: 'pn[0]',        attr: '9' },
  {bits:  9,  name: 'pn[1]',        attr: '9'},
  {bits:  9,  name: 'pn[2]',        attr: '9'},
```

chapter4.adoc
rsahita: I am reverting this change since this spa is an input into the lookup process and those upper bits may or may not be zero. HW will still fault if they are not, but HW does not force them to be 0 as inputs into the lookup process. similar convention is followed for Sv as well.

chapter4.adoc
```diff
@@ -1,390 +1,432 @@
...
-and _mpte.info_ that are reserved for future standard use, per
-<<Smmpt-rw-l2-encoding>>, are set within _mpte_, stop and raise an access-fault
-exception corresponding to the access type.
+Such MPTEs behave identically to leaf MPTEs in the MPT access permission lookup
+process described in <<MPT_ACC_LKUP>>, except that:
 
-7. The _mpte_ is a valid leaf _mpte_. Fetch the access-permissions for the
-physical address per the steps described below:
+* For RV32, the `N` bit may be set to 1 in the level 0 leaf MPTEs.
+* For RV64, the `N` bit may be set to 1 in the level 0 and level 1 leaf MPTEs.
```
rsahita, line 359:
```suggestion
* For RV64, the `N` bit may be set to 1 in the leaf MPTEs.
```

chapter4.adoc
rsahita: non-0 benefit to validation and verification.

SiFiveHolland: The only required change on the hardware side is not forcing an access fault in the case L=1 and N=1.

On the other hand, making the extension optional adds cost (to development, validation, and verification) to any software that wants to use leaf NAPOT pages. I would argue that the software cost outweighs the hardware validation cost. Optional extensions aren't free.

chapter4.adoc
rsahita: This can be achieved with one additional 4K page and an entry on that page. I don't think we should add a mode for that.

chapter4.adoc
rsahita: I am ok making SmmptNapot a requirement for the Smmpt modes.

chapter4.adoc
rsahita: ack. I will describe it as when the entry at i > 0 is the leaf, the bits pa.pn[i-1] are included in the range offset.

chapter4.adoc
SiFiveHolland: There's also the performance overhead of the extra level of MPT walk. Small/embedded/real-time systems are less likely to have MPTE caches, so the performance hit will be more noticeable.

chapter4.adoc
SiFiveHolland: Your update looks good to me.

chapter4.adoc
```diff
@@ -1,390 +1,445 @@
...
+[width="100%",cols="10%,30%,30%,30%",options="header",]
+|===
+| *_i_* a| *_mpte.PPN[i]_*  a| *Description*        a|*_mpte.napot_bits_*
+| < 2   a| x xxxx xxxx xxxx a| Reserved             a| -
+|   2   a| x xxx1 0000 0000 a| 2 MiB contiguous L2  a| 9
+|   2   a| x xxx0 xxxx xxxx a| Reserved             a| -
+|   3   a| x xxx1 0000 0000 a| 2 MiB L3             a| 9
+|   3   a| x xxx0 xxxx xxxx a| Reserved             a| -
+|   4   a| x 1000 0000 0000 a| 16 MiB contiguous L4 a| 12
+|   4   a| x 0xxx xxxx xxxx a| Reserved             a| -
```
SiFiveHolland, line 414:
```suggestion
| *_i_* a| *_mpte.PPN[0]_* a| *Description*        a| *_mpte.napot_bits_*
|    0  a|     x xxxx xxxx a| Reserved             a| -
|  > 0  a|     x xxxx xxx1 a| Reserved             a| -
|  > 0  a|     x xxxx xx1x a| Reserved             a| -
|  > 0  a|     x xxxx x1xx a| Reserved             a| -
|  > 0  a|     x xxxx 1xxx a| Reserved             a| -
|  > 0  a|     x xxx1 xxxx a| Reserved             a| -
|  > 0  a|     x xx1x xxxx a| Reserved             a| -
|  > 0  a|     x x1xx xxxx a| Reserved             a| -
|  > 0  a|     x 1xxx xxxx a| Reserved             a| -
|  > 0  a|     1 0000 0000 a| 2 MiB next-level MPT a| 9
|  > 0  a|     0 xxxx xxxx a| Reserved             a| -
```

* The first row indicates that non-leaf NAPOT MPTEs are disallowed in the level 0 (leaf) MPT.
* The middle rows indicate that NAPOT sizes smaller than 2 MiB are reserved.
* The second to last row is for the allowed encoding that points to a 2 MiB *next* level table. The size of the table containing the non-leaf NAPOT MPTE only matters in that there must be space for 512 MPTEs. This is always the case for the RV64 Smmpt modes, so no other conditions need to be stated.
* The last row indicates that NAPOT sizes greater than 2 MiB are reserved.

If you really want a Smmtt64 L4 (root) MPTE to contain a single set of NAPOT MPTEs, instead of 8, you would need another table/row to indicate that for _i_ = 4, the combination `mpte.PPN[1] = x xxxx x100` and `mpte.PPN[0] = 0 0000 0000` is allowed. But as I mentioned on #126, I don't think that's necessary.

chapter4.adoc
```diff
@@ -1,390 +1,445 @@
...
-batch optimized via the use of the sequence `SFENCE.W.INVAL`, `MINVAL.SPA` and
-`SFENCE.INVAL.IR`.
+Such MPTEs behave identically to non-leaf MPTEs in the MPT access permission
+lookup process described in <<MPT-lookup>>, except that:
+
+ * If the encoding in _mpte_ is valid according to table <<Smmpt-napot>>,
+   then instead of returning the original value of the _mpte_, implicit read
+   of a non-leaf NAPOT _mpte_ returns a copy of `mpte` in which
+   _mpte.ppn[i][mpte.napot_bits-1:0]_ are replaced by
+   _pn[i][mpte.napot_bits-1:0]_. If the encoding in _mpte_ is reserved
```
SiFiveHolland, line 424:
```suggestion
   _mpte.ppn[mpte.napot_bits-1:0]_ are replaced by
   _pa.pn[i][mpte.napot_bits-1:0]_. If the encoding in _mpte_ is reserved
```

>> It is always mpte.ppn[0] that is overwritten, not mpte.ppn[i]
> 
> That does not seem right

OK, I think the confusion is because _mpte.ppn_ is not split into fields for each MPT level -- it's just one 52 bit field. This is different from Svnapot, where _pte.ppn_ is split into 9-bit subfields.

For clarity, apologies if this is obvious: A non-leaf NAPOT MPTE points to a 2 MiB page (21 bit offset) made up of 512 x 4 KiB pages (9 bit page number + 12 bit page offset) It's these 9 bits that vary based on which of the NAPOT MPTEs is being traversed -- bits 20:12 of the _next_ MPTE's physical address, or in other words _mpte.ppn[8:0]_. Regardless of which level MPT contains the NAPOT MPTE, it's always the same bits that vary. Consider that if _mpte.ppn[17:9]_ were overwritten, the next-level MPT would not be contiguous! It would be 512 4 KiB pages, each 2 MiB apart from each other, covering a total of 1 GiB of address space.

The way Svnapot is written _looks_ like it would extend to i > 0, but does not. Non-leaf Svnapot would similarly need to always replace _pte.ppn[0]_, not the higher bits.

chapter4.adoc
```diff
@@ -1,390 +1,445 @@
...
+   according to <<Smmpt-napot>> then an access-fault exception corresponding to
+   the original access type must be raised.
+ * Implicit reads of non-leaf NAPOT MPTEs may create MPT walk cache entries
+   mapping _a_ + _j_*PTESIZE to a copy of _mpte_ in which
+   _mpte[i][pte.napot_bits-1:0]_ is replaced by _pn[i][mpte.napot_bits-1:0]_,
+   for any and all _j_ such that _j_ >> _napot_bits_ = _pn[i].napot_bits_.
```
SiFiveHolland, line 430:
```suggestion
   _mpte[mpte.napot_bits-1:0]_ is replaced by _pa.pn[i][mpte.napot_bits-1:0]_,
   for any and all _j_ such that _j_ >> _napot_bits_ = _pa.pn[i]_ >> _napot_bits_.
```

* Same fix as above
* Copy/paste issues from Svnapot

chapter4.adoc
```diff
@@ -1,390 +1,445 @@
...
 
 [NOTE]
 ====
-MPT lookups that began while `mmpt` was active are not required to complete or
-terminate when `mmpt` is no longer active, unless a `MFENCE.SPA` instruction
-matches the `SDID` (and optionally, `PADDR`) is executed. The `MFENCE.SPA`
-instruction must be used to ensure that updates to the `MPT` data structures are
-observed by subsequent implicit reads to those structures by a hart.
+Some implementations may cache non-leaf MPTE to accelerate the MPT access-type
```
SiFiveHolland, line 434: editorial:

```suggestion
Some implementations may cache non-leaf MPTEs to accelerate the MPT access-type
```

SiFiveHolland (commented):
Here's the promised fixes for Smmptnlnapot

chapter4.adoc
rsahita, line 414:
```suggestion
[[Smmpt-napot]]
.Non-leaf `MPTE` encodings when _mpte_.N=1
[width="100%",cols="10%,30%,30%,30%",options="header",]
|===
| *_i_* a| *_mpte.PPN[8:0]_* a| *Description* a| *_mpte.napot_bits_*
|    0  a|       x xxxx xxxx a| Reserved      a| -
|  > 0  a|       x 1xxx xxxx a| Reserved      a| -
|  > 0  a|       1 0000 0000 a| 2 MiB contiguous next-level MPT a| 9
|  > 0  a|       0 xxxx xxxx a| Reserved      a| -
```

rsahita: we could just state it this way then

rsahita: see below for shortened table

chapter4.adoc
rsahita, line 430:
```suggestion
   _mpte.ppn[mpte.napot_bits-1:0]_ is replaced by _pa.pn[i][mpte.napot_bits-1:0]_,
   for any and all _j_ such that _j_ >> _napot_bits_ = _pa.pn[i]_ >> _napot_bits_.
```

rsahita: fixed a typoe _mpte -> _mpte.ppn below

chapter4.adoc
```diff
@@ -1,390 +1,442 @@
...
-====
+[[Smmpt-napot]]
+.Non-leaf `MPTE` encodings when _mpte_.N=1
+[width="100%",cols="10%,30%,30%,30%",options="header",]
+|===
+| *_i_* a| *_mpte.PPN[8:0]_* a| *Description* a| *_mpte.napot_bits_*
+|    0  a|       x xxxx xxxx a| Reserved      a| -
+|  > 0  a|       x 1xxx xxxx a| Reserved      a| -
+|  > 0  a|       1 0000 0000 a| 2 MiB contiguous next-level MPT a| 9
+|  > 0  a|       0 xxxx xxxx a| Reserved      a| -
```
rsahita, line 411:
```suggestion
| *_i_* a| *_mpte.PPN[8:0]_* a| *Description* a| *_mpte.napot_bits_*
|    0  a|       x xxxx xxxx a| Reserved      a| -
|  > 0  a|       x 1xxx xxxx a| Reserved      a| -
|  > 0  a|       x x1xx xxxx a| Reserved      a| -
|  > 0  a|       x Xx1x xxxx a| Reserved      a| -
|  > 0  a|       x xxx1 xxxx a| Reserved      a| -
|  > 0  a|       x xxxx 1xxx a| Reserved      a| -
|  > 0  a|       x xxxx x1xx a| Reserved      a| -
|  > 0  a|       x xxxx xx1x a| Reserved      a| -
|  > 0  a|       x xxxx xxx1 a| Reserved      a| -
|  > 0  a|       1 0000 0000 a| 2 MiB contiguous next-level MPT a| 9
|  > 0  a|       0 xxxx xxxx a| Reserved      a| -
```

chapter4.adoc
```diff
@@ -1,390 +1,449 @@
...
+|  > 0  a|       x 1xxx xxxx a| Reserved      a| -
+|  > 0  a|       x x1xx xxxx a| Reserved      a| -
+|  > 0  a|       x Xx1x xxxx a| Reserved      a| -
+|  > 0  a|       x xxx1 xxxx a| Reserved      a| -
+|  > 0  a|       x xxxx 1xxx a| Reserved      a| -
+|  > 0  a|       x xxxx x1xx a| Reserved      a| -
+|  > 0  a|       x xxxx xx1x a| Reserved      a| -
+|  > 0  a|       x xxxx xxx1 a| Reserved      a| -
+|  > 0  a|       1 0000 0000 a| 2 MiB contiguous next-level MPT a| 9
+|  > 0  a|       0 xxxx xxxx a| Reserved      a| -
```
rsahita, line 418:
```suggestion
| *_i_* a| *_mpte.PPN[8:0]_* a| *Description* a| *_mpte.napot_bits_*
|    0  a|       x xxxx xxxx a| Reserved      a| -
|  > 0  a|       x xxxx xxx1 a| Reserved      a| -
|  > 0  a|       x xxxx xx1x a| Reserved      a| -
|  > 0  a|       x xxxx x1xx a| Reserved      a| -
|  > 0  a|       x xxxx 1xxx a| Reserved      a| -
|  > 0  a|       x xxx1 xxxx a| Reserved      a| -
|  > 0  a|       x xx1x xxxx a| Reserved      a| -
|  > 0  a|       x x1xx xxxx a| Reserved      a| -
|  > 0  a|       x 1xxx xxxx a| Reserved      a| -
|  > 0  a|       1 0000 0000 a| 2 MiB contiguous next-level MPT a| 9
|  > 0  a|       0 xxxx xxxx a| Reserved      a| -
```

SiFiveHolland (approved):
LGTM! Thanks for your additional fix.

chapter4.adoc
SiFiveHolland: It looks like this fix didn't get applied -- only the second bullet got changed.

PR #126, opened 2025-02-22
Simplify the definition of Smmptnlnapot

SiFiveHolland:
I found the definition of Smmptnlnapot rather difficult to understand, and I think there are some inconsistencies in the current text (e.g. a non-leaf NAPOT MPTE always points to a 2 MiB page, because the size depends on the _next_ MPT level, and only the root MPT is ever larger than 4 KiB).

I've reformulated the definition in a way that I believe is equivalent but much easier to understand. With this formulation, I don't think any of the extra text about MPT walk caches is necessary, because there are no synthetic values anymore. The hardware would see the cached non-leaf NAPOT MPTE and repeat the same modified algorithm. There just wouldn't even be a cache entry for the skipped level.

### DIFF
--- a/chapter4.adoc
+++ b/chapter4.adoc
@@ -375,55 +375,23 @@ or Smmpt64 extensions must be implemented.
 
 === Smmptnlnapot - NAPOT Memory Protection Table
 
-The Smmptnlnapot extension supports NAPOT non-leaf MPTEs. This extension is
-RV64 only, and depends on one of Smmpt43, Smmpt52 or Smmpt64.
+The Smmptnlnapot extension supports NAPOT non-leaf MPTEs. This extension
+applies to RV64 only, and depends on one of Smmpt43, Smmpt52, or Smmpt64. It
+allows skipping an intermediate level of the MPT lookup process by coalescing
+512 adjacent MPT pages into a single 2 MiB superpage. When a NAPOT non-leaf
+MPTE is encountered, the next two levels of the MPT lookup process (each
+normally computing an MPTE address using 9 bits of the physical address) are
+replaced by a single level which computes the MPTE address using the next 18
+bits of the physical address.
 
-For RV64, when the `MPTE.L`=0 and `MPTE.N`=1, the non-leaf `MPTE`
-is part of a range of MPTEs at that level of MPT with the same value for
-the L, N, PPN, and V bits.
-
-The following Non-leaf `MPTE` encodings are defined when _mpte_.N is 1.
-
-[[Smmpt-napot]]
-.Non-leaf `MPTE` encodings when _mpte_.N=1
-[width="100%",cols="10%,30%,30%,30%",options="header",]
-|===
-| *_i_* a| *_mpte.PPN[i]_*  a| *Description*        a|*_mpte.napot_bits_*
-| < 2   a| x xxxx xxxx xxxx a| Reserved             a| -
-|   2   a| x xxx1 0000 0000 a| 2 MiB contiguous L2  a| 9
-|   2   a| x xxx0 xxxx xxxx a| Reserved             a| -
-|   3   a| x xxx1 0000 0000 a| 2 MiB L3             a| 9
-|   3   a| x xxx0 xxxx xxxx a| Reserved             a| -
-|   4   a| x 1000 0000 0000 a| 16 MiB contiguous L4 a| 12
-|   4   a| x 0xxx xxxx xxxx a| Reserved             a| -
-|===
+When `MPTE.L`=0 and `MPTE.N`=1, the MPTE is a NAPOT non-leaf MPTE. `MPTE.PPN`
+points to a naturally-aligned 2 MiB superpage, so `MPTE.PPN`[8:0] are reserved
+and must be zero.
 
 Such MPTEs behave identically to non-leaf MPTEs in the MPT access permission
-lookup process described in <<MPT-lookup>>, except that:
+lookup process described in <<MPT_ACC_LKUP>>, except that:
 
- * If the encoding in _mpte_ is valid according to table <<Smmpt-napot>>,
-   then instead of returning the original value of the _mpte_, implicit read
-   of a non-leaf NAPOT _mpte_ returns a copy of `mpte` in which
-   _mpte.ppn[i][mpte.napot_bits-1:0]_ are replaced by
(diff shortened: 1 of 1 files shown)


ved-rivos:
While the only smnlnapot code initially provided is for 9 napot bits, the extension will form the base for future extensions that may want additional encodings

SiFiveHolland:
I can remove the sentence about `MPTE.PPN`[8:0] and replace it with a table similar to what was there before, using the previous encoding with bit 8 set. Given that we are not defining the behavior of the other reserved encodings, is my formulation of the MPT walk behavior still acceptable?

ved-rivos:
The property of Smnlnapot is that it can be completely ignored by an implementation - even when it support Smnlnapot - and continue to process the MPTE as-if it was not a napot PTE. The next two bullets are providing a formula for computing addresses and it closely mirrors the Svnapot formulation which is familiar and well understood by people familiar with the ISA. Formulating it other ways than Svnapot may lead to people needing to understand an otherwise identical concept using two different styles of specification.

SiFiveHolland:
> The property of Smnlnapot is that it can be completely ignored by an implementation - even when it support Smnlnapot - and continue to process the MPTE as-if it was not a napot PTE.

Are you thinking of Smmptnapot (leaf NAPOT MPTEs)? Even in the original formulation, it is not possible to process a non-leaf NAPOT MPTE as if it was a non-leaf non-NAPOT MPTE. MPTE.PPN doesn't actually point to the next table level. The NAPOT encoding forces bit 8 to be set, while the next table level is at an address with bit 8 clear (because it is naturally aligned).

ved-rivos:
Ignore for the purposes of caching MPTEs in the mpt-walk-cache i.e. an implementation may choose to cache such as MPTE as one, two, or up to 512 standard MPTEs using content generated from the napot nl-MPTE.

chapter4.adoc
```diff
@@ -375,55 +375,23 @@ or Smmpt64 extensions must be implemented.
...
-|   2   a| x xxx1 0000 0000 a| 2 MiB contiguous L2  a| 9
-|   2   a| x xxx0 xxxx xxxx a| Reserved             a| -
-|   3   a| x xxx1 0000 0000 a| 2 MiB L3             a| 9
-|   3   a| x xxx0 xxxx xxxx a| Reserved             a| -
-|   4   a| x 1000 0000 0000 a| 16 MiB contiguous L4 a| 12
-|   4   a| x 0xxx xxxx xxxx a| Reserved             a| -
-|===
+The Smmptnlnapot extension supports NAPOT non-leaf MPTEs. This extension
+applies to RV64 only, and depends on one of Smmpt43, Smmpt52, or Smmpt64. It
+allows skipping an intermediate level of the MPT lookup process by coalescing
```
rsahita, line 380: It does not allow skipping an intermediate level per se - the caching takes advantage of it, but the entries still need to be specified.

chapter4.adoc
```diff
@@ -375,55 +375,23 @@ or Smmpt64 extensions must be implemented.
...
-MPT walk cache pressure in some scenarios. Simpler implementations may simply
-create a single entry in their page walk cache using the transformed mpte value
-returned by the implicit access to the mpte.
-
-Depending on the need, the NAPOT scheme may be extended to other levels of the
-MPT in the future.
-====
+ * In step 4, subtract 2 from _i_ instead of 1.
+ * In step 2 on the following iteration, the `MPTE` address is calculated as
+   _a_ + _pa.pn[i+1:i]_ x MPTESIZE.
```
rsahita, line 396: It does not allow skipping an intermediate level per se - the caching takes advantage of it, but the entries still need to be specified. so the subtract 2 is not right.

IMO, we should keep this section as currently specified as it also mirrors the language used for Svnapot that people are familiar with - as written it also makes it clear that other encodings may be defined in the future if needed. Also I think the implementation note is required to explain the caching advantages.

rsahita:
IMO, we should keep this section as currently specified as it also mirrors the language used for Svnapot that people are familiar with - as written it also makes it clear that other encodings may be defined in the future if needed. Also I think the implementation note is required to explain the caching advantages. (More comments inline in the proposed text).

SiFiveHolland:
In that case, I will abandon this change and instead submit some fixes to the existing text:
* It is always _mpte.ppn[0]_ that is overwritten, not _mpte.ppn[i]_
* Off-by-one errors in the table (now that the table levels are 0-indexed)
* Smmpt64's root MPT need not be treated specially; it can contain 8 consecutive sets of 512 non-leaf NAPOT MPTEs, if desired. This would simplify hardware implementations.

chapter4.adoc
SiFiveHolland: Yes, this is true. My formulation accomplishes the same result without relying so heavily on the MPT walk cache. (We were asked to target embedded systems, after all.)

rsahita:
sg. so should we close this PR?

reg these items:
>It is always mpte.ppn[0] that is overwritten, not mpte.ppn[i]
That does not seem right

>Off-by-one errors in the table (now that the table levels are 0-indexed)
The table seems right since we have p[4], p[3], p[2] for nlnapot

>Smmpt64's root MPT need not be treated specially; it can contain 8 consecutive sets of 512 non-leaf NAPOT MPTEs, if desired. This would simplify hardware implementations.
In any case 12 bits will need to be considered as napot_bits

SiFiveHolland:
> sg. so should we close this PR?
> 
> reg these items:
> 
> > It is always mpte.ppn[0] that is overwritten, not mpte.ppn[i]
> 
> That does not seem right

I explain this in my comment on #121 

> > Off-by-one errors in the table (now that the table levels are 0-indexed)
> 
> The table seems right since we have p[4], p[3], p[2] for nlnapot

Non-leaf NAPOT entries are also valid when i == 1, pointing to a 2 MiB contiguous L0 table.

> > Smmpt64's root MPT need not be treated specially; it can contain 8 consecutive sets of 512 non-leaf NAPOT MPTEs, if desired. This would simplify hardware implementations.
> 
> In any case 12 bits will need to be considered as napot_bits

napot_bits == 9 works fine. Your 4096-entry root MPT would include:

* Entries 0-511 -> one non-leaf NAPOT group
* Entries 512-1023 -> a second non-leaf NAPOT group
* Entries 1024-1535 -> a third non-leaf NAPOT group
* etc.

And these 8 NAPOT groups could point to consecutive pages if you want. There's no requirement that the entire root MPT has to be a _single_ NAPOT group.

PR #127, opened 2025-02-27
Additional Smmptnlnapot editorial fixes

SiFiveHolland:
I missed these issues when redoing the review of Smmptnlnapot.

 * Fix broken section link
 * Remove redundant "table"
 * Fix wording in references to mpte.ppn and pa.pn

I don't think these will cause any problems for ARC review.


rsahita (approved):
thanks

PR #129, opened 2025-03-20
Address issue 122

rsahita:


Baixyzz:
> cc @Baixyzz

Hi, @rsahita . I noticed that the reference to mip.LSDEIP and sip.LSDEIP being read-only was removed. However, I think there is still ambiguity here.

> The mip.LSDEIP bit is 1 if and only if the bitwise logical AND of CSRs msdeip and msdeie is nonzero in any bit. 

Does this mean that the mip.LSDEIP bit is writable but can only be written to 0?

rsahita:
no such restriction - it is writable normally. I updated the PR as well, and merged - ptal.

PR #130, opened 2025-03-23
Update to address ARC feedback on Smsdia

ved-rivos:


chapter7.adoc
```diff
@@ -181,21 +264,21 @@ which aliases bits 63:32 of `msdeip`. When MXLEN=64, `msdeiph` does not exist.
...
-that APLIC.
+When the supervisor interrupt domain identified by __i__ is implemented by an
+APLIC, the bit __i__ indicates the state of the S-level external interrupt
+pending signal provided by the supervisor interrupt domain in that APLIC.
 
-When the interrupt controller identified by __i__ is an IMSIC, the bit __i__
-indicates the logical OR of the S-level and all VS-level external interrupt
-pending signals provided by that IMSIC.
+When the supervisor interrupt domain identified by __i__ is implemented by an
+IMSIC, the bit __i__ indicates the logical OR interrupt signals from all the
```
rsahita, line 274:
```suggestion
IMSIC, the bit __i__ indicates the logical OR of the interrupt signals from all the
```

rsahita (approved):
reviewing latest updates

gagachang:
The preview PDF downloaded from https://github.com/riscv/riscv-smmtt/pull/130/checks shows the wrong formats in section 6.2:

image

chapter7.adoc
```diff
@@ -254,3 +333,74 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+supervisor interrupt domain are indicated in the `hgeip` CSR if the bit
+corresponding to those guest interrupt files in `mgirfd` is 1 else they are
+indicated as 0.
+
+If bit of `mgirfd` selected by `hstatus.VGEIN` is 1 and the supervisor interrupt
+domain selected by `msdcfg.SIDN` is an implemented supervisor interrupt domain,
+and is not zero, and is an IMSIC, then the `vsiselect` and `vstopei` CSRs operate
+on the registers of the corresponding guest interrupt file.
+
+If bit of `mgirfd` selected by `hstatus.VGEIN` is 1 and the supervisor interrupt
```
gagachang, line 399:
```diff
- If bit of `mgirfd` selected by `hstatus.VGEIN` is 1
+ If bit of `mgirfd` selected by `hstatus.VGEIN` is 0
```

gagachang:
I have some comments for `Smirfd` extension.

Current Linux implementation probes number of implemented guess external interrupt files by `hgeie` CSR and device tree.
See: https://github.com/torvalds/linux/blob/master/arch/riscv/kvm/aia.c#L648

Does `mgirfd` also affect the read/write behavior of `hgeie` ?
Current privileged spec defines `hgeie` as :
```
- The hgeie register is an HSXLEN-bit read/write register ...
- The least-significant bits are implemented first, apart from bit 0.
```
IIUC, `Smirfd` will make `hgeie` not an HSXLEN-bit "read/write" register.
If bit N of `mgirfd` is not 1, the bit N of `hgeie` will not be writable.
Thus `Smirfd` changes `hgeie` as an "WARL" register.
Also, `Smirfd` breaks "The least-significant bits are implemented first" rule, and the writable bits in `hgeie` would not be contiguous bits.


Does `mgirfd` also affect the WLRL behavior of `hstatus.VGEIN` ?
What's the behavior of a hypervisor writes `N` into `hstatus.VGEIN` when bit N of `mgirfd` is not 1 ?

chapter7.adoc
```diff
@@ -25,77 +25,158 @@ to the RDSM. The RDSM, in the course of handling an external interrupt, should
...
+[latexmath]
+++++
+\left( (2^{sn} - 1) \times 2^I \right) \mathbin{\&} B = 0
+++++
+
+This ensures that the address expression:
+
+[latexmath]
+++++
+s \times 2^I + g \times 2^E + B + h \times 2^F
```
rsahita, line 106:
```suggestion
s \times 2^I + g \times 2^E + B + h \times 2^D
```

chapter7.adoc
```diff
@@ -25,77 +25,158 @@ to the RDSM. The RDSM, in the course of handling an external interrupt, should
...
+++++
+
+is equivalent to the bitwise composition:
+
+[latexmath]
+++++
+s \times 2^I \,|\, g \times 2^E \,|\, B \,|\, h \times 2^F
```
rsahita, line 113:
```suggestion
s \times 2^I \,|\, g \times 2^E \,|\, B \,|\, h \times 2^D
```

chapter7.adoc
```diff
@@ -254,3 +333,74 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+
+The `S` bit when set to 1 enables access to the supervisor-level interrupt file
+in the supervisor interrupt domain selected by `msdcfg.SIDN`. Each bit of the
+`GIRF` field when set to 1 enables access to the Guest interrupt file in the
+supervisor interrupt domain selected by `msdcfg.SIDN`. If `GEILEN` is nonzero,
+bit `GEILEN:0` shall be writeable in `mgirfd`, and all other bit positions shall
+be read-only zeros.
+
+If the `S` bit of `mgirfd` is 1 and the supervisor interrupt domain selected by
+`msdcfg.SIDN` is an implemented supervisor interrupt domain, and is not zero,
```
rsahita, line 371:
```suggestion
`msdcfg.SIDN` is an implemented supervisor interrupt domain,
```

chapter7.adoc
```diff
@@ -254,3 +333,74 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+If the `S` bit of `mgirfd` is 0 then:
+
+* The S-level external interrupt pending signal indicated in `mip.SEIP` is 0.
+* All non-custom values of `siselect` that access IMSIC registers designate an
+  _inaccessible_ register. When `siselect` holds the number of an _inaccessible_
+  register, attempts from M-mode or HS-mode to access `sireg` raise an illegal
+  instruction exception.
+
+If supervisor interrupt domain selected by `msdcfg.SIDN` is an implemented
+supervisor interrupt domain, and is not zero, and is an IMSIC, then the VS-level
```
rsahita, line 388:
```suggestion
supervisor interrupt domain, and is an IMSIC, then the VS-level
```

chapter7.adoc
```diff
@@ -254,3 +333,74 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+external interrupt pending signals of the guest interrupt files of selected
+supervisor interrupt domain are indicated in the `hgeip` CSR if the bit
+corresponding to those guest interrupt files in `mgirfd` is 1 else they are
+indicated as 0.
+
+If bit of `mgirfd` selected by `hstatus.VGEIN` is 1 and the supervisor interrupt
+domain selected by `msdcfg.SIDN` is an implemented supervisor interrupt domain,
+and is not zero, and is an IMSIC, then the `vsiselect` and `vstopei` CSRs operate
```
rsahita, line 396:
```suggestion
and is an IMSIC, then the `vsiselect` and `vstopei` CSRs operate
```

chapter7.adoc
```diff
@@ -254,3 +333,74 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+on the registers of the corresponding guest interrupt file.
+
+If bit of `mgirfd` selected by `hstatus.VGEIN` is 1 and the supervisor interrupt
+domain selected by `msdcfg.SIDN` is an implemented supervisor interrupt domain,
+and is not zero, and is an IMSIC, then the `vsiselect` and `vstopei` CSRs operate
```
rsahita, line 401:
```suggestion
and is an IMSIC, then the `vsiselect` and `vstopei` CSRs operate
```

rsahita (approved):

rsahita (approved):
LGTM - thanks

chapter7.adoc
```diff
@@ -254,3 +332,102 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+If the supervisor interrupt domain selected by `msdcfg.SIDN` is an implemented
+supervisor interrupt domain, then:
+
+* The VS-level external interrupt pending signals of the guest interrupt files
+  of the selected supervisor interrupt domain are indicated in the `hgeip` CSR
+  if the bit corresponding to their guest external interrupt number in `mgirfd`
+  is 1 else they are indicated as 0.
+
+* Bits `GEILEN:1` of `hgeie` are writable if the bit at same position in in
+  `mgirfd` is 1. All other bits are read-only zeroes.
```
rsahita, line 393:
```suggestion
* Bits `GEILEN:1` of `hgeie` are writable if the bit at same position in
  `mgirfd` is 1. All other bits are read-only zeroes.
```

ved-rivos:
> 
> Does `mgirfd` also affect the read/write behavior of `hgeie` ? Current privileged spec defines `hgeie` as :

Of the `GEILEN:1` bits in `hgeie`, only the bits where the corresponding bit in `mgirfd` is 1 are writable. All other bits are read-only zeros.

> IIUC, `Smirfd` will make `hgeie` not an HSXLEN-bit "read/write" register. If bit N of `mgirfd` is not 1, the bit N of `hgeie` will not be writable. (Am I correct ?) Thus `Smirfd` changes `hgeie` as an "WARL" register. Also, `Smirfd` breaks "The least-significant bits are implemented first" rule, and the writable bits in `hgeie` would not be contiguous bits.

The register will stay HSXLEN-bit wide. It also remains WARL i.e. only the bits where `mgirfd` has 1 are legal and writable. The writable bits in `hgeie` may not be contigous. Likewise the bits in `hgeip` may not be contigous. The memory layout of the register files may also not be contiguous. The Smirfd extension is intended to be used by enlightened supervisor domains. Here typically there is a donor domain that donates a interrupt file to recipient domain using SBI functions provided by RDSM - for example, see the [COVE Interrupt Extension](https://github.com/riscv-non-isa/riscv-ap-tee/blob/6cc1337d90d37e4ac3d655365106bb91700eb4f4/src/sbi_cove.adoc?plain=1#L1672) to SBI.


> Does `mgirfd` also affect the WLRL behavior of `hstatus.VGEIN` ? What's the behavior of a hypervisor writes `N` into `hstatus.VGEIN` when bit N of `mgirfd` is not 1 ?

The behaviour is same as selecting an unimplemented guest interrupt file.

ved-rivos:
> @ved-rivos Thanks for added clarification!
> 
> BTW, would it be better to change the name `Smirfd` to `Smirfdeleg` or something else suffixed with `deleg` ? RISC-V has several extensions related to "delegation" and their name are suffixed with `deleg`:
> 
> * `Smcdeleg` for supervisor counter delegation
> * `Smtdeleg` for trigger delegation

That is a good idea. I will change the name to Smirfdeleg. Thanks for the suggestion

chapter7.adoc
```diff
@@ -254,3 +332,109 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+`msdcfg.SIDN` is implemented then:
+
+* The S-level external interrupt pending signal of supervisor-level interrupt
+  file of the selected _supervisor interrupt domain_ is indicated in `mip.SEIP`.
+* The `siselect`, `sireg`, and `stopei` CSRs operate on the registers of the
+  supervisor-level interrupt register file in the selected _supervisor interrupt
+  domain_.
+
+If the `S` bit of `mirfd` is 0 or the supervisor domain selected by
+`msdcfg.SIDN` is not an implemented then:
```
rsahita, line 378:
```suggestion
If the `S` bit of `mirfd` is 0 or the supervisor interrupt domain selected by
`msdcfg.SIDN` is not implemented then:
```

chapter7.adoc
```diff
@@ -254,3 +332,109 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+* The S-level external interrupt pending signal indicated in `mip.SEIP` is 0.
+* All non-custom values of `siselect` that access IMSIC registers designate an
+  _inaccessible_ register. When `siselect` holds the number of an _inaccessible_
+  register, attempts from M-mode or S/HS-mode to access `sireg` raise an illegal
+  instruction exception.
+* Access to `stopei` from M-mode or S/HS-mode raises an illegal instruction
+  exception.
+
+If the hypervisor extension is also implemented and the _supervisor interrupt
+domain_ selected by `msdcfg.SIDN` is an implemented then:
```
rsahita, line 389:
```suggestion
If the hypervisor extension is also implemented and the _supervisor interrupt
domain_ selected by `msdcfg.SIDN` is implemented then:
```

chapter7.adoc
```diff
@@ -254,3 +332,109 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+
+* If `hstatus.VGEIN` is not 0 and the bit selected by `VGEIN` in `mirfd` is 1
+  then the `vsiselect`, `vsireg`, and `vstopei` CSRs operate on the registers of
+  the _guest interrupt file_ selected by `VGEIN`. If `VGEIN` is 0 or the the bit
+  selected by `VGEIN` in `mirfd` is 0 then the `vsiselect`, `vsireg`, and
+  `vstopei` CSRs operate as-if the `VGEIN` is the number of an unimplemented
+  guest external interrupt.
+
+If the hypervisor extension is also implemented and the _supervisor interrupt
+domain_ selected by `msdcfg.SIDN` is not an implemented then:
```
rsahita, line 407:
```suggestion
If the hypervisor extension is also implemented and the _supervisor interrupt
domain_ selected by `msdcfg.SIDN` is not implemented then:
```

rsahita:
PR was reviewed in TG on 4/8 and was kept open for the week to allow for additional feedback.
Merging since all review comments were addressed.

chapter7.adoc
```diff
@@ -254,3 +332,109 @@ delegation supports optimizing the exit sequence from such supervisor domain by
...
+* The S-level external interrupt pending signal of supervisor-level interrupt
+  file of the selected _supervisor interrupt domain_ is indicated in `mip.SEIP`.
+* The `siselect`, `sireg`, and `stopei` CSRs operate on the registers of the
+  supervisor-level interrupt register file in the selected _supervisor interrupt
+  domain_.
+
+If the `S` bit of `mirfd` is 0 or the _supervisor interrupt domain_ selected by
+`msdcfg.SIDN` is not implemented then:
+
+* The S-level external interrupt pending signal indicated in `mip.SEIP` is 0.
```
SiFiveHolland, line 380: Just to confirm, is my understanding correct here?

When a bit is clear in `mirfd`, and the interrupt signal for that interrupt file does not affect `mip.SEIP`/`hgeip`, the interrupt signal for that interrupt file _still does_ affect the `msdeip` bit for that SIDN. So the way to check for interrupts from other supervisor domains sharing the current SIDN is to check for `msdeip[SIDN] AND (NOT mip.SEIP) AND (NOT mip.SGEIP)`.

This means the RDSM will need to keep the `msdeie` bit for the current SIDN set at all times. But as long as LSDEI is configured to have a lower priority than SEI/SGEI, and `mideleg[SEI]` is set, an interrupt for the currently active supervisor domain will not trap to M-mode, because SEI/SGEI and LSDEI will be asserted simultaneously, and similarly clearing SEI/SGEI will also deassert LSDEI at the same time.

ved-rivos: Not quite. I think we are missing a section here on the interaction of `mirfd` with `LSDEI`.

gagachang: > So the way to check for interrupts from other supervisor domains sharing the current SIDN is to check for `msdeip[SIDN] AND (NOT mip.SEIP) AND (NOT mip.SGEIP)`.

Which component needs to check for such interrupts in this case? The donor domain?

SiFiveHolland: Whichever component makes the decision to preempt a supervisor domain, which could be the RDSM, or a privileged supervisor domain if LSDEI is delegated.

gagachang:
I would like to describe my understanding based on CoVE TVM model.
Please correct me if I am wrong. Thanks!

For a hart supports Smirfdeleg.
That's say Host SD0 and Confidential SD1 share IMSIC supervisor interrupt domain 1 (SIDN=1).
Thus, when RDSM switches to both SD0 and SD1 it will set `msdcfg.SIDN`=1.

Host SD0 owns all IMSIC interrupt files.
Thus, when RDSM switches to host SD0, `mirfd.S` and `mirfd[GEILEN:1]` will be set as all 1.
Now, host SD0 delegates(donates) the guest interrupt files 1 to Confidential SD1 to be used by TVM.
When RDSM switches to confidential SD1, it will set `mirfd.S`=1 and `mirfd[1]`=1.

RDSM can further set mideleg.SEIP=1 when switching to confidential SD1. So that host SD0 external interrupts would be indicated as `mip.SEIP` during confidential SD1 execution. Confidential SD1 then yields the execution to host SD0 to let it handle the pending interrupts.

@SiFiveHolland  So the way to check for interrupts from other supervisor domains sharing the current SIDN:
- If current domain is host SD0, it can check `hgeip` for interrupts of TVM in confidential SD1(TVM uses guest interrupt file)
- If current domain is confidential SD1, it can check `sip.SEIP` for interrupts of host SD0

SiFiveHolland:
> Now, host SD0 delegates(donates) the guest interrupt file 1 to the confidential SD1 to be used by TVM. When RDSM switches to the confidential SD1, it will set `mirfd.S`=1 and `mirfd[1]`=1.

This doesn't sound correct to me. `mirfd.S` controls access to SD0's interrupt file, which SD1 should not be able to access. And `mirfd[1]` controls SD1's interrupt file, which SD0 should not be able to access. So I would expect:

* When running SD0, RDSM sets `mirfd[GEILEN:2]`=1, `mirfd[1]`=0, and `mirfd.S`=1.
* When running SD1, RDSM sets `mirfd[GEILEN:2]`=0, `mirfd[1]`=1, and `mirfd.S`=0.

> RDSM can further set `mideleg.SEIP`=1 when switching to the confidential SD1. So that host SD0 external interrupts would be indicated as `mip.SEIP` during confidential SD1 execution. Confidential SD1 then yields the execution to the host SD0 to let it handle the pending interrupts.
> 
> @SiFiveHolland So the way to check for interrupts from other supervisor domains sharing the current SIDN:
> 
> * If current domain is host SD0, it can check `hgeip` for interrupts of TVM in confidential SD1 (TVM uses guest interrupt file)

If `mirfd[1]`=0, then `hgeip[1]`=0 as well, per the text of Smirfdeleg that was just merged:

```
  * The VS-level external interrupt pending signals of the _guest interrupt files_
  of the selected _supervisor interrupt domain_ are indicated in the `hgeip` CSR
  if the bit corresponding to their guest external interrupt number in `mirfd`
  is 1 else they are indicated as 0.
```

> * If current domain is confidential SD1, it can check `sip.SEIP` for interrupts of host SD0

Similarly, if `mirfd.S` is cleared, `mip.SEIP` will be zero (unless set manually by software).

So I don't see any way to check the pending status of the other interrupt files except LSDEI, unless we change Smirfdeleg to (optionally) report the pending status of the inaccessible interrupt files.

chapter7.adoc
ved-rivos: As I mentioned there is a section missing on the interaction of `mirfd` and `LSDEI`. For the active `SIDN`, only the interrupt pending signals from the inaccessible interrupt files contribute to `LSDEI`. This is pending update.

SiFiveHolland: Does that imply, for consistency, that on systems without Smirfdeleg, none of the interrupt pending signals for the currently active SIDN contribute to LSDEI (since all interrupt files in that supervisor interrupt domain are accessible)?

gagachang: So LSDEI consists of pending signals from other supervisor interrupt domains and inaccessible interrupt files?

gagachang:
> This doesn't sound correct to me. `mirfd.S` controls access to SD0's interrupt file, which SD1 should not be able to access. And `mirfd[1]` controls SD1's interrupt file, which SD0 should not be able to access. So I would expect:
> 
> * When running SD0, RDSM sets `mirfd[GEILEN:2]`=1, `mirfd[1]`=0, and `mirfd.S`=1.
> * When running SD1, RDSM sets `mirfd[GEILEN:2]`=0, `mirfd[1]`=1, and `mirfd.S`=0.

Have you checked the note at the end of section 6.4 ?
```
The Smirfdeleg is intended to be used to delegate interrupt files to enlightened
(a.k.a. para-virtualized) supervisor domains. Usually, one of the supervisor
domains participating in the delegation is a donor domain that initially has access
to all the interrupt files.
```
IIUC, `Smirfdeleg` is designed for CoVE TVM model. @ved-rivos also mentioned [COVE Interrupt Extension](https://github.com/riscv-non-isa/riscv-ap-tee/blob/6cc1337d90d37e4ac3d655365106bb91700eb4f4/src/sbi_cove.adoc?plain=1#L1672) in my previous comments.

However, you mentioned "a privileged supervisor domain" (like secure SD vs. non-secure SD ?).
I am not sure if `Smirfdeleg` is applicable in this scenario.
Maybe "two dedicated supervisor interrupt domains" is better in this scenario ?

SiFiveHolland:
> > This doesn't sound correct to me. `mirfd.S` controls access to SD0's interrupt file, which SD1 should not be able to access. And `mirfd[1]` controls SD1's interrupt file, which SD0 should not be able to access. So I would expect:
> > 
> > * When running SD0, RDSM sets `mirfd[GEILEN:2]`=1, `mirfd[1]`=0, and `mirfd.S`=1.
> > * When running SD1, RDSM sets `mirfd[GEILEN:2]`=0, `mirfd[1]`=1, and `mirfd.S`=0.
> 
> Have you checked the note at the end of section 6.4 ?
> 
> ```
> The Smirfdeleg is intended to be used to delegate interrupt files to enlightened
> (a.k.a. para-virtualized) supervisor domains. Usually, one of the supervisor
> domains participating in the delegation is a donor domain that initially has access
> to all the interrupt files.
> ```
> 
> IIUC, `Smirfdeleg` is designed for CoVE TVM model. @ved-rivos also mentioned [COVE Interrupt Extension](https://github.com/riscv-non-isa/riscv-ap-tee/blob/6cc1337d90d37e4ac3d655365106bb91700eb4f4/src/sbi_cove.adoc?plain=1#L1672) in my previous comments.

My interpretation of the note (and my understanding of the CoVE model), is that the non-confidential supervisor domain (SD0 in the example) starts out with access to all of the interrupt files (all bits in `mirfd` set). But when it calls `sbi_covi_convert_aia_imsic()`, it loses access to the corresponding interrupt file (a bit in `mirfd` would be cleared while executing SD0). And similarly, the bit in `mirfd` would be set again after calling `sbi_covi_reclaim_tvm_aia_imsic()`. SD0 is not in the TCB of SD1, so it should not be able to access SD1's interrupt file.

> However, you mentioned "a privileged supervisor domain" (like secure SD vs. non-secure SD ?). I am not sure if `Smirfdeleg` is applicable in this scenario. Maybe "two dedicated supervisor interrupt domains" is better in this scenario ? Because both the two SDs need its own supervisor-level interrupt file.

That comment was only about LSDEI delegation -- an example of where the RDSM may not be the entity making scheduling decisions based on interrupt pending signals. And yes, I was referring to a secure/non-secure (i.e. GlobalPlatforms TEE) model. Smirfdeleg could be used in this model, as long as at least one of the OSes (TEE OS or REE OS) was enlightened and could work with `mirfd.S` cleared.

gagachang:
> My interpretation of the note (and my understanding of the CoVE model), is that the non-confidential supervisor domain (SD0 in the example) starts out with access to all of the interrupt files (all bits in `mirfd` set). This is why the note says "initially". But when it calls `sbi_covi_convert_aia_imsic()`, it loses access to the corresponding interrupt file (a bit in `mirfd` would be cleared while executing SD0). And similarly, the bit in `mirfd` would be set again after calling `sbi_covi_reclaim_tvm_aia_imsic()`. SD0 is not in the TCB of SD1, so it should not be able to access SD1's interrupt file.

Oh, it makes sense. I didn't notice the word "initially".
Maybe we need a programming guide somewhere for API sequence and `mirfd` ...

> That comment was only about LSDEI delegation -- an example of where the RDSM may not be the entity making scheduling decisions based on interrupt pending signals. And yes, I was referring to a secure/non-secure (i.e. GlobalPlatforms TEE) model. Smirfdeleg could be used in this model, as long as at least one of the OSes (TEE OS or REE OS) was enlightened and could work with `mirfd.S` cleared.

Yeah, if one of the OSes doesn't need supervisor-level interrupt domain then Smirfdeleg is applicable.

chapter7.adoc
```diff
@@ -25,77 +25,157 @@ to the RDSM. The RDSM, in the course of handling an external interrupt, should
...
-identical.
-
-Upon scheduling a supervisor domain for execution on a hart, the RDSM requires a
-mechanism to appoint the linked interrupt controller as the active one for the
-hart. Selecting an interrupt controller also chooses the linked S and VS-level
-interrupt pending signals as those detected by the hart for instigating the
-associated traps. When CSRs are utilized to interface with an interrupt
-controller's state, the interaction is with the active interrupt controller.
+If `CI` is the Child Index from the `source[i]` corresponding to an interrupt
+source `_i_` then the MSI target address is thus computed as follows:
```
gagachang, line 153: Hi @ved-rivos 
The name of register is `sourcecfg[i]` not `source[i]`.

IIUC, the `sourcecfg[i]` here means the `sourcecfg[i]` in parent machine-level interrupt domain, instead of the `sourcecfg[i]` in child supervisor-level interrupt domain.
Would it be better to explicitly state this here for better readability?

gagachang: > If `CI` is the Child Index from the ...

Besides, why using the word "If" here ?
Using "If" means there are other interpretations of `CI`.
Does the spec intend to do this ?

ved-rivos: > IIUC, the sourcecfg[i] here means the sourcecfg[i] in parent machine-level interrupt domain, instead of the sourcecfg[i] in child supervisor-level interrupt domain.

The parent of any supervisor-level interrupt domain is always a machine-level domain.

ved-rivos: > Besides, why using the word "If" here ?

The word was used to establish the variable `CI`. Will reword it if it is confusing.

chapter7.adoc
ved-rivos: Yes.

ved-rivos: Please review https://github.com/riscv/riscv-smmtt/pull/136

PR #151, opened 2025-06-09
Updates to rv64 mmpt and mpte per ARC feedback

rsahita:
Updates to rv64 mmpt and mpte per ARC feedback

### DIFF
--- a/chapter4.adoc
+++ b/chapter4.adoc
@@ -57,9 +57,9 @@ MPTE entry in <<rv32-mpte-leaf>>.
 {reg: [
   {bits:  1,  name: 'V'},
   {bits:  1,  name: 'L=0'},
-  {bits:  8,  name: 'RESERVED'},
+  {bits:  8,  name: 'Reserved'},
   {bits:  22, name: 'PPN'},
-], config:{lanes: 2, hspace:1024}}
+], config:{lanes: 1, hspace:1024}}
 ....
 
 [caption="Figure {counter:image}: ", reftext="Figure {image}"]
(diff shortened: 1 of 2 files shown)


chapter4.adoc
```diff
@@ -234,9 +234,9 @@ range determined by that MPTE.
 {reg: [
   {bits:   1, name: 'V'},
   {bits:   1, name: 'L=0'},
-  {bits:   8, name: 'RESERVED'},
-  {bits:  52, name: 'PPN'},
-  {bits:   1, name: 'RSV'},
+  {bits:   8, name: 'Reserved'},
+  {bits:  44, name: 'PPN'},
```
SiFiveHolland, line 238: Can you clarify why the size of this field was changed for all RV64 MPT modes? According to the text, they have the same MPTE format. If the intention was to restrict the PPN range to the number of physical address bits supported by a given MPT mode, then:
 * Smmpt43 needs a 31 bit PPN
 * Smmpt52 needs a 40 bit PPN
 * Smmtt64 needs a 52 bit PPN

none of which is 44 bits.

ved-rivos: 44 represents the PPn width of the maximum physical address that can be addressed by the virtual memory system

SiFiveHolland: Why is the virtual memory system relevant here? There is no need for MPTs to be accessible via virtual memory translation, as these tables are never accessed by S-mode. M-mode does not use translation, so it could access any valid physical address up to 2^64-1. In fact, if a system has valid physical addresses >2^56, it would be ideal to place MPTs there, so as to maximize the physical memory available for S-mode virtual memory translation.

If the expectation is that no systems will have valid physical addresses >2^56, since those would be untranslatable via virtual memory, why do we have Smmtt64 and not the simpler/more regular Smmtt61 (4k root MPT)?

PR #152, opened 2025-06-09
Editorial updates to Smsdia

ved-rivos:


chapter7.adoc
```diff
@@ -414,12 +419,27 @@ domain_ selected by `msdcfg.SIDN` is not implemented then:
 When `Smifdeleg` is implemented, `hip.SGEIP` is 1 if and only if the bitwise
 logical-AND of CSRs `hgeip`, `hgeie`, and `mifd` is nonzero in any bit.
```
rsahita, line 420: this rule is repeated below - maybe this one is not needed?

PR #173, opened 2025-07-31
ARC review updates to Smsdid

ved-rivos:
- msdcfg should be MXLEN and not 32-bit
- update to split SSM to SSRM and SSMM


rsahita (commented):
lgtm

PR #306, opened 2026-08-24
Fix the looks of nmrmcfg and msdcfg and make it build

radimkrcmar:
Before:
image

image

After:
image

image

And it also builds in docker; hopefully in other environments as well.


Makefile
```diff
@@ -14,7 +14,7 @@
 
 DATE ?= $(shell date +%Y-%m-%d)
 VERSION ?= 0.9.0
-REVMARK ?= Draft - Frozen for ARC review
+REVMARK ?= 'Draft - Frozen for ARC review'
```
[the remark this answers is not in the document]
radimkrcmar: Implemented the suggestion.

Makefile
```diff
@@ -25,9 +25,9 @@ ASCIIDOCTOR_PDF := asciidoctor-pdf
 OPTIONS := --trace \
            -a compress \
            -a mathematical-format=svg \
-           -a revnumber=${VERSION} \
-           -a revremark=${REVMARK} \
-           -a revdate=${DATE} \
+           -a revnumber="${VERSION}" \
+           -a revremark="${REVMARK}" \
+           -a revdate="${DATE}" \
```
gagachang, line 30: I encountered this revmark build failure today.
It's weird that the failure only happens in local build.
CI workflow works well without fixing ...
https://github.com/riscv/riscv-smmtt/actions/runs/32934257829/job/98072349347

sameo (approved):

Makefile
radimkrcmar: `.github/workflows/build-pdf.yml` adds `REVMARK=Draft` to the environment, which overrides the default and doesn't contain any spaces to trip this bug.

radimkrcmar:
Rebased and added the required signatures.  No changes to code.

PR #313, opened 2026-09-17
Clarify the IOMPT table walk SQRID, RCID, MCID

andrewdellow:
add text to clarify the IOMPT table walk SQRID, RCID, MCID are those associated with the SD.


radimkrcmar (approved):
LGTM.

Btw. have we decided what to do with the Bare IO-MPT mode?
1. There would be no complication if we removed Bare.
2. We could add iompt_rmcfg register for Bare transactions based on mnrmcfg.
3. a) If we keep Bare in current shape, we could specify that Bare mode sets implementation-defined QRID and takes RCID and MCID from IOMMU.
b) If we go back to previous state, we could say that QRID, RCID, and MCID in Bare mode are implementation-specific.
5. We could extend the Bare selector with QRID field and take RCID and MCID from IOMMU.

andrewdellow:
> LGTM.
> 
> Btw. have we decided what to do with the Bare IO-MPT mode?
> 
> 1. There would be no complication if we removed Bare.
> 2. We could add iompt_rmcfg register for Bare transactions based on mnrmcfg.
> 3. a) If we keep Bare in current shape, we could specify that Bare mode sets implementation-defined QRID and takes RCID and MCID from IOMMU.
>    b) If we go back to previous state, we could say that QRID, RCID, and MCID in Bare mode are implementation-specific.
> 4. We could extend the Bare selector with QRID field and take RCID and MCID from IOMMU.

in bare mode there are no MPT walks, and there is a default IOMMU instance, so i would expect the SQRID/RCID/ MCID mechanism to still apply ? i.e. in bare mode it is assumed there is only a single SD, and the default IOMMU is associated with that SD , and configured appropriately, and any invalid device context will be dropped.

radimkrcmar:
> in bare mode there are no MPT walks, and there is a default IOMMU instance, so i would expect the SQRID/RCID/ MCID mechanism to still apply ?

Yes, all transactions on a platform that supports Supervisor Domains and CBQRI carry QRID, RCID, and MCID.
IOMMU provides a register to configure the global default for RCID and MCID in `iommu_qosid` register and overrides in `DC.ta`.

I think we should close the gap of unspecified SQRID in IO-MPT Bare mode.
(My previous message covered the options that I consider to be on the Pareto front.)

andrewdellow:
> > in bare mode there are no MPT walks, and there is a default IOMMU instance, so i would expect the SQRID/RCID/ MCID mechanism to still apply ?
> 
> Yes, all transactions on a platform that supports Supervisor Domains and CBQRI carry QRID, RCID, and MCID. IOMMU provides a register to configure the global default for RCID and MCID in `iommu_qosid` register and overrides in `DC.ta`.
> 
> I think we should close the gap of unspecified SQRID in IO-MPT Bare mode. (My previous message covered the options that I consider to be on the Pareto front.)

yes, agree, without the IOMPT, even if there was a 'known, default' SDID there appears to be no source of SQRID.  would you propose adding BARE_SQRID to the IOMPT control register ?

radimkrcmar:
Honestly, I don't understand the goal of Bare mode anymore, but the new register seems as the best solution if we want Bare mode to be more than a convenience for early platform boot.
(At the cost of making it even more inconvenient there.)

Values from the new register would take the place of SDCFG entries when computing QRID and the effective RCID/MCID.
Implementations without Bare wouldn't need to support the configuration, since IO-MPT doesn't do any implicit accesses.

When talking about "register", there are two main options -- either add it as a top-level register and have
```
bare_qosid <- sqrid
wait until status.busy is 0
control <- Bare | iommu_id
```

Or add it to the command interface, which I think is more in line with On mode and would also allows us to pack Bare mode iommu configuration:
```
data1 <- sqrid | iommu_id
command <- SET_BARE
wait until status.busy is 0
control <- Bare
```

For comparison, this is how to achieve the same behavior with On mode:
```
data1 <- Bare
data2 <- sqrid
command <- SET_SDCFG_ENTRY | sdid
wait until status.busy is 0
data1 <- sdid | iommu_id | match_all | exclude_tee
command <- SET_SDLC_ENTRY | 0
wait until status.busy is 0
control <- On
```
(Which shows that On still has the ability to allow TEE and achieve better transparency/composability than Bare -- maybe a gap we want to address too.)

andrewdellow:
> Honestly, I don't understand the goal of Bare mode anymore, but the new register seems as the best solution if we want Bare mode to be more than a convenience for early platform boot. (At the cost of making it even more inconvenient there.)
> 
> Values from the new register would take the place of SDCFG entries when computing QRID and the effective RCID/MCID. Implementations without Bare wouldn't need to support the configuration, since IO-MPT doesn't do any implicit accesses.
> 
> When talking about "register", there are two main options -- either add it as a top-level register and have
> 
> ```
> bare_qosid <- sqrid
> wait until status.busy is 0
> control <- Bare | iommu_id
> ```
> 
> Or add it to the command interface, which I think is more in line with On mode and would also allows us to pack Bare mode iommu configuration:
> 
> ```
> data1 <- sqrid | iommu_id
> command <- SET_BARE
> wait until status.busy is 0
> control <- Bare
> ```
> 
> For comparison, this is how to achieve the same behavior with On mode:
> 
> ```
> data1 <- Bare
> data2 <- sqrid
> command <- SET_SDCFG_ENTRY | sdid
> wait until status.busy is 0
> data1 <- sdid | iommu_id | match_all | exclude_tee
> command <- SET_SDLC_ENTRY | 0
> wait until status.busy is 0
> control <- On
> ```
> 
> (Which shows that On still has the ability to allow TEE and achieve better transparency/composability than Bare -- maybe a gap we want to address too.)

I was expecting it to be a field in control.   i.e 
 control <- Bare | BARE_IOMMU_ID | BARE_QRID

also is MODE=Bare permitted when the implementation does not contain an IOMMU source for RCID and MCID? I assume it doesn't make much sense so expect that in the case of no IOMMU support, RCID and MCID would be UNDEFINED.

radimkrcmar:
> I was expecting it to be a field in control. i.e
> control <- Bare | BARE_IOMMU_ID | BARE_QRID

Sorry, misunderstood that as case 2.  Case 4 loses control over SSRM, SSMM, SLR, SML, S-RCID, and S-MCID, which simplifies the platform boot, but loses the parity with what M-mode on a hart can and must do with a Bare domain.

That could be completely fine, so Bare IO-MPT sets SSRM/SSMM=0 and SLR/SML=-1, making S-RCID and S-MCID irrelevant in the effective RCID/MCID computation.

> also is MODE=Bare permitted when the implementation does not contain an IOMMU source for RCID and MCID? I assume it doesn't make much sense so expect that in the case of no IOMMU support, RCID and MCID would be UNDEFINED.

As I read the specification, IO-MPT requires at least one IOMMU_ID and we require that the IOMMU_ID provides RCID/MCID.

(How the IOMMU_ID provides the RCID/MCID is unspecified, so the platform could just as well hardwire the values in IO-MPT if there is only a single IOMMU_ID that cannot change the ids).

andrewdellow:
I put in a very simply PR to add a 4 bit BARE_QRID, with the expectation that RCID and MCID come from the default IOMMU instance selected by BARE_IOMMU_ID.  please comment !

chapter6.adoc
```diff
@@ -440,8 +441,11 @@ and the following rules apply:
   not TEE-associated and to abort those that are TEE-associated.
 
 * For allowed transactions, the I/O Bridge shall use the IOMMU instance identified by 
-  `control.BARE_IOMMU_ID` to perform address translation. If the implementation does not 
-  include an IOMMU, `BARE_IOMMU_ID` is read-only zero and address translation is not performed. 
+  `control.BARE_IOMMU_ID` to perform address translation. The `QRID` associated with the transaction 
+  shall be the value held in `control.BARE_QRID`. If the implementation does not 
+  include an IOMMU, `BARE_IOMMU_ID` is read-only zero, address translation is not performed, and the 
+  `RCID` and `MCID` associated with transactions processed while `MODE` is `Bare` are `UNSPECIFIED`.
```
radimkrcmar, line 447: I'd avoid the last sentence.  The rest of the specification implies that BARE_IOMMU_ID=0 is the first instance of an IOMMU or something that acts as an IOMMU for the purposes of IO-MPT.
Then we specify here that in Bare mode, the effective RCID and MCID are the RCID and MCID values provided by IOMMU.

chapter6.adoc
```diff
@@ -440,8 +441,11 @@ and the following rules apply:
   not TEE-associated and to abort those that are TEE-associated.
 
 * For allowed transactions, the I/O Bridge shall use the IOMMU instance identified by 
-  `control.BARE_IOMMU_ID` to perform address translation. If the implementation does not 
-  include an IOMMU, `BARE_IOMMU_ID` is read-only zero and address translation is not performed. 
+  `control.BARE_IOMMU_ID` to perform address translation. The `QRID` associated with the transaction 
+  shall be the value held in `control.BARE_QRID`. If the implementation does not
```
radimkrcmar, line 445: We're using `QRID` here and `effective SQRID` below, which is a bit confusing.  I didn't figure out why we're using `SQRID` in the SDCFG fields, so converting everything `QRID` seems more consistent.

chapter6.adoc
andrewdellow: i was trying to avoid abiguity if the implementation doesnt have any IOMMU - then there is still no source for RCID and MCID, but QRID there is.

chapter6.adoc
andrewdellow: on reading, the SQRID is sort of the QRID associated with an SD.  QRID is the actual identifier, so in a way it makes sense to distinguish SQRID when SDs are active and just use QRID when there are no SDs ?  i'm open to changing it, alternatively maybe adding a note to say SQRID holds a QRID value associated with requests originating from a SD ?

chapter6.adoc
radimkrcmar: There is one more place that could be clarified then: the SDCFG rule uses RCID-value and MCID-value from IOMMU, even if there is no IOMMU (making the concept of IOMMU abstract).  This led me to infer that the values would be always be provided to IO-MPT, possibly hardwired, but adding a partial clarification here makes the SDCFG case more ambiguous.

Also, the "address translation is not performed" is not explicitly required when mode is On -- I think it would be best to have the same semantics for `Bare` and `On` modes.

(And if we wanted Bare to be even more aligned with On mode, the BARE_IOMMU_ID would be ignored on write, and unspecified on read when there isn't an associated IOMMU.)

andrewdellow:
further clarifications after discussion.  please review...

